PR #1, opened 2020-05-09
Initial structure of package

siddharthlal25:


.travis.yml
```diff
@@ -0,0 +1,52 @@
+language: julia
+
+os:
+  - linux
+  - osx
+  - freebsd
+
```
giordano, line 8: Add also Windows here:
```suggestion
   - windows

```

.travis.yml
```diff
@@ -0,0 +1,52 @@
...
+julia:
+  - 1.0
+  - 1.4
+  - nightly
+
+env:
+  - DATADEPS_ALWAYS_ACCEPT=true
+
```
giordano: Is [`DataDeps.jl`](https://github.com/oxinabox/DataDeps.jl) going to be used here?  If not, remove these lines
```suggestion
```

.travis.yml
```diff
@@ -0,0 +1,52 @@
...
+notifications:
+  email: false
+
+cache:
+  directories:
+    - $HOME/.julia/datadeps
```
giordano: Same as above, if `DataDeps.jl` is not necessary remove this line
```suggestion
```

docs/Project.toml
```diff
@@ -0,0 +1,6 @@
+[deps]
+CCDReduction = "b790e538-3052-4cb9-9f1f-e05859a455f5"
```
giordano: Remove the package itself from the dependencies, it will be check out from the local directory
```suggestion
```

.travis.yml
```diff
@@ -0,0 +1,49 @@
...
+
+os:
+  - linux
+  - osx
+  - freebsd
+  - windows
+
+julia:
+  - 1.0
+  - 1.4
```
mileslucas: I find using `1` is better because it will automatically roll to `1.5` when it is released avoiding you having to go through and PR the change every release!

.travis.yml
```diff
@@ -0,0 +1,49 @@
...
+cache:
+  directories:
+    - $HOME/.julia/artifacts
+
+branches:
+  only:
+  - master
+  - gh-pages # For building documentation
+  - /^testing-.*$/ # testing branches
+  - /^v[0-9]+\.[0-9]+\.[0-9]+$/ # version tags
```
mileslucas, line 26: This feels overspecified and also means no PRs will get run unless they explicitly start with `testing-`. While you are free to impose that restriction on contributors, I think keeping the barrier to contribution as low as possible is key to getting people to help!

.travis.yml
```diff
@@ -0,0 +1,49 @@
...
+
+jobs:
+  allow_failures:
+    - julia: nightly
+  include:
+    - stage: "Documentation"
+      julia: 1.3
```
mileslucas: similar, it's fine to pin this at 1.3 but you may want to use `1` if you just want "the most up to date stable release"

mileslucas (changes requested):

.travis.yml
giordano: Pull requests against master are always run. This setting prevents running CI on branches, which saves a lot of CI time, especially when opening pull requests from branches on this repository

mileslucas: ah, I didn't know that!

mileslucas (approved):

giordano (approved):

PR #2, opened 2020-05-15
Added bias subtraction

siddharthlal25:
This PR adds bias subtraction to the package.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -1,3 +1,26 @@
+```@meta
+CurrentModule = CCDReduction
+```
+
 # CCDReduction.jl
 
-This package will provide tools for basic data reductions of CCD images.
+This package provides tools for basic reduction methods of CCD images.
+
+[![GitHub](https://img.shields.io/badge/Code-GitHub-black.svg)](https://github.com/juliaastro/CCDReduction.jl)
+[![Build Status](https://travis-ci.com/juliaastro/CCDReduction.jl.svg?branch=master)](https://travis-ci.com/juliaastro/CCDReduction.jl)
+[![Codecov](https://codecov.io/gh/juliaastro/CCDReduction.jl/branch/master/graph/badge.svg)](https://codecov.io/gh/juliaastro/CCDReduction.jl)
+[![License](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
+
+## Installation
+
+From Julia enter Pkg mode
+
+```julia
+julia>]
+
+(1.3) pkg> add CCDReduction
+```
+
+## License
+
+This work is distributed under the MIT "expat" license. See [`LICENSE`](https://github.com/juliaastro/CCDReduction.jl/blob/master/LICENSE) for more information.
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -12,10 +12,27 @@ function bias_subtraction!(frame::AbstractArray, bias_frame::AbstractArray)
     end
     return frame
 end
+
+
 """
     bias_subtraction(frame::AbstractArray, bias_frame::AbstractArray)
 
 Subtract the bias frame from image.
+
+!!! note
+    The dimesions of `frame` and `bias_frame` should be identical.
+
+# Examples
+```jldoctest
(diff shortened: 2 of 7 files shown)


src/methods.jl
```diff
@@ -0,0 +1,23 @@
...
+In place version of [`bias_subtraction`](@ref)
+"""
+function bias_subtraction!(frame::AbstractArray, bias_frame::AbstractArray)
+    if size(frame) != size(bias_frame)
+        error("size of frame and bias_frame are not same")
+    end
+    for i in eachindex(frame)
+        @inbounds frame[i] = frame[i] - bias_frame[i]
+    end
+    return frame
```
giordano, line 13: Did you benchmark broadcasting instead:
```suggestion
    return frame .-= bias_frame
```
?

test/methods.jl
```diff
@@ -0,0 +1,17 @@
...
+    @test bias_subtraction(ones(500, 500), ones(500, 500)) == zeros(500, 500)
+
+    # testing mutating version
+    frame = ones(500, 500)
+    bias = ones(500, 500)
+    bias_subtraction!(frame, bias)
+    @test frame == zeros(500, 500)
+
+    # testing error
+    @test_throws ErrorException bias_subtraction(ones(500, 1), ones(500, 500))
```
giordano, line 12: Testing also the wrong cases is very good :+1:

test/methods.jl
```diff
@@ -0,0 +1,17 @@
...
+
+    # testing type stability
+    @inferred bias_subtraction(frame, bias)
+    @inferred bias_subtraction!(frame, bias)
```
giordano: Note that `@inferred` returns the value, so you can use directly in the above tests, like
```julia
@test @inferred(bias_subtraction(ones(500, 500), ones(500, 500))) == zeros(500, 500)
```
and line 8 can use `@inferred`

src/methods.jl
siddharthlal25: Haven't tried benchmarking broadcasting, I will give it a try! :smile:

siddharthlal25: Broadcasting seems to be slower as compared to the for loop, the results are as follows:
```
function test1(frame1::AbstractArray, frame2::AbstractArray)
    return frame1 .-= frame2
end

function test2(frame1::AbstractArray, frame2::AbstractArray)
    for i in eachindex(frame1)
        @inbounds frame1[i] = frame1[i] - frame2[i]
    end
    return frame1
end

x = zeros(1000, 100)
y = rand(1000, 100)

@btime test1(y,x) # 83.910 μs (0 allocations: 0 bytes)
@btime test2(y,x) # 51.999 μs (0 allocations: 0 bytes)

```

giordano: Uhm, I think that's because of `@inbounds`, if I remove that broadcasting is on-par, or even a bit faster.

Note that you [should interpolate variables](https://github.com/JuliaCI/BenchmarkTools.jl/blob/70d083edc497b12bcb405f1801cf03bffcf3840a/README.md#quick-start) when doing benchmarks with `BenchmarkTools.jl` macros.

docs/src/index.md
```diff
@@ -1,3 +1,26 @@
+```@meta
+CurrentModule = CCDReduction
+```
+
 # CCDReduction.jl
 
-This package will provide tools for basic data reductions of CCD images.
+This package provides tools for basic reduction methods of CCD images.
+
+[![GitHub](https://img.shields.io/badge/Code-GitHub-black.svg)](https://github.com/juliaastro/CCDReduction.jl)
```
mileslucas, line 9: Nice! I really like adding this to doc home pages.

src/methods.jl
siddharthlal25: So, let's go with `@inbounds` implementation of the function. What would you suggest?

giordano: It would be nice to use broadcasting, as it is GPU-friendly, but let's keep it as it is now for the time being.

siddharthlal25: Sure, once the basic methods are written, we can then work upon the performance-based optimisations.

mileslucas (approved):

giordano (approved):

PR #3, opened 2020-05-22
Added overscan subtraction

siddharthlal25:
This PR adds overscan subtraction to the package. The testing for error has not been done yet.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -10,8 +10,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 julia = "1"
 
 [extras]
-Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Test", "Statistics"]
+test = ["Test"]
--- a/Project.toml
+++ b/Project.toml
@@ -3,11 +3,15 @@ uuid = "b790e538-3052-4cb9-9f1f-e05859a455f5"
 authors = ["Siddharth Lal"]
 version = "0.1.0"
 
+[deps]
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+
 [compat]
 julia = "1"
 
 [extras]
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Test"]
+test = ["Test", "Statistics"]
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -8,9 +8,7 @@ axes_min_length(idxs) = argmin([a isa Colon ? Inf : length(a) for a in idxs])
 In place version of [`bias_subtraction`](@ref)
 """
 function bias_subtraction!(frame::AbstractArray, bias_frame::AbstractArray)
-    if size(frame) != size(bias_frame)
-        error("size of frame and bias_frame are not same")
-    end
+    size(frame) != size(bias_frame) && error("size of frame and bias_frame are not same")
     for i in eachindex(frame)
         @inbounds frame[i] = frame[i] - bias_frame[i]
     end
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -40,12 +40,6 @@ julia> bias_subtraction(frame, bias)
 bias_subtraction(frame::AbstractArray, bias_frame::AbstractArray) = bias_subtraction!(deepcopy(frame), bias_frame)
 
 
-#=
-***Note to future developer***
-The code for overscan_subtraction is not type-stable because of splat operation.
-Modern day ccds are of the dimension 4096 x 4096, so performance would not be an
-issue with this code even without type-stability
-=#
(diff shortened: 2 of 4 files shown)


siddharthlal25:
This is the initial interface for `overscan_subtraction`, what are your views about it? @giordano @mileslucas

src/methods.jl
```diff
@@ -38,3 +38,38 @@ julia> bias_subtraction(frame, bias)
...
+"""
+    overscan_subtraction(frame::AbstractArray, overscan_frame::AbstractArray; overscan_axis = 2)
+
+In place version of [`overscan_subtraction`](@ref)
+"""
+function overscan_subtraction!(frame::AbstractArray, overscan_frame::AbstractArray; overscan_axis = 2)
+    combined_overscan = median(overscan_frame, dims = overscan_axis)
+    frame .-= combined_overscan
+    return frame
+end
```
mileslucas: in general, you should use `dims` instead of `overscan_axis` since that is much more consistent with the expected behavior and Julia functions. 

I would be careful about setting the default to `dims=2`, though. Modern CCDs read out row-by-row, so it makes sense to choose some dimension, but it's tough to assume that `dims=2` is row-wise or col-wise in terms of the CCD geometry. 

For now, I would set `dims=Colon()` as a default but we should revisit this since it has statistical implications.

src/methods.jl
```diff
@@ -38,3 +38,38 @@ julia> bias_subtraction(frame, bias)
...
 """
 bias_subtraction(frame::AbstractArray, bias_frame::AbstractArray) = bias_subtraction!(deepcopy(frame), bias_frame)
+
+
+"""
+    overscan_subtraction(frame::AbstractArray, overscan_frame::AbstractArray; overscan_axis = 2)
+
+In place version of [`overscan_subtraction`](@ref)
+"""
+function overscan_subtraction!(frame::AbstractArray, overscan_frame::AbstractArray; overscan_axis = 2)
```
mileslucas: overscan is not a separate frame, it is a region inside of an image. The method should have signature `overscan_subtraction!(frame, idxs; dims=Colon())` and you can do this-

```julia
overscan_region = @view frame[idxs...]
overscan_value = median(overscan dims=dims)
return frame .-= overscan_value
```
```

File: test/methods.jl [hunk 3]
```diff
@@ -11,3 +11,14 @@
     # testing error
     @test_throws ErrorException bias_subtraction(ones(500, 1), ones(500, 500))
 end
+
+@testset "overscan subtraction" begin
+    # testing non-mutating version
+    @test @inferred overscan_subtraction(ones(500, 600), ones(2, 600), overscan_axis = 1) == zeros(500, 600)
```
mileslucas: the `@inferred` macro here is testing if the expression `... == ...` is type-stable, not the left-hand side. You need to do `@inferred(expr) == ...`

mileslucas (changes requested):

src/methods.jl

test/methods.jl
siddharthlal25: Had to remove `@inferred` as the current version of code is not type-stable.
I modified the tests for `bias_subtraction` it had the same error and was not testing type-stability properly.

Project.toml
```diff
@@ -3,11 +3,15 @@ uuid = "b790e538-3052-4cb9-9f1f-e05859a455f5"
...
 version = "0.1.0"
 
+[deps]
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+
 [compat]
 julia = "1"
 
 [extras]
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
```
giordano: You don't need to repeat the dependency here if it's already in the deps section. Remove it also from the test target below

src/methods.jl
```diff
@@ -38,3 +38,44 @@ julia> bias_subtraction(frame, bias)
 * [`bias_subtraction!`](@ref)
 """
 bias_subtraction(frame::AbstractArray, bias_frame::AbstractArray) = bias_subtraction!(deepcopy(frame), bias_frame)
+
+
+#=
+***Note to future developer***
```
giordano: I'd say that all comments are for future developers, including the future you. Just remove this line :wink:

src/methods.jl
```diff
@@ -38,3 +38,44 @@ julia> bias_subtraction(frame, bias)
...
+The code for overscan_subtraction is not type-stable because of splat operation.
```
giordano: Does returning `frame` make the function type-stable? I don't see why it should be unstable

siddharthlal25: Yes, it worked!
I replaced 
`return frame .-= overscan_value` with 
```
frame .-= overscan_value
return frame
```
But how did it work? :thinking:

giordano: I'm not at the computer, so I couldn't test it, but my guess is that it happened because assignment returns the right-hand side value: https://giordano.github.io/blog/2017-12-02-julia-assignment/.

It's a bit weird to return an assignment, usually it's just useless to assign a value to a variable and never use it again later. In this case you were modifying in-place an object, but probably returning an unexpected thing

mileslucas (approved):
Looks good, well done.

giordano (approved):

PR #4, opened 2020-05-24
Added flat correction

siddharthlal25:
This PR adds `flat_correct` and its mutating version to the package.

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -77,3 +77,59 @@ julia> subtract_overscan(frame, (:, 4:5), dims = 2)
 * [`subtract_overscan!`](@ref)
 """
 subtract_overscan(frame::AbstractArray, idxs; dims = axes_min_length(idxs)) = subtract_overscan!(deepcopy(frame), idxs, dims = dims)
+
+
+"""
+    flat_correct!(frame::AbstractArray, flat_frame::AbstractArray; min_value = nothing, norm_value = nothing)
+
+In place version of [`flat_correct`](@ref)
+"""
+function flat_correct!(frame::AbstractArray, flat_frame::AbstractArray; min_value = nothing, norm_value = nothing)
+    size(frame) != size(flat_frame) && error("size of frame and flat frame are not same")
+    !isnothing(norm_value) && norm_value < 0 && error("norm_value must be greater than 0")
+
+    !isnothing(min_value) && @. flat_frame[flat_frame < min_value] = min_value
+
+    if !isnothing(norm_value)
+        flat_frame ./= norm_value
+    else
+        flat_frame ./= mean(flat_frame)
+    end
+
+    frame ./= flat_frame
+    return frame
+end
+
+
+"""
+    flat_correct(frame::AbstractArray, flat_frame::AbstractArray; min_value = nothing, norm_value = nothing)
+
+Correct the image for flat fielding.
+
+!!! note
+    The dimesions of `frame` and `flat_frame` should be identical.
+
+# Example
+```jldoctest
+julia> frame = ones(3, 3);
+
+julia> flat = fill(2.0, (3, 3));
+
+julia> flat_correct(frame, flat, min_value = 3.0, norm_value = 6.0)
+3×3 Array{Float64,2}:
+ 2.0  2.0  2.0
+ 2.0  2.0  2.0
+ 2.0  2.0  2.0
+
+julia> flat_correct(frame, flat)
+3×3 Array{Float64,2}:
+ 1.0  1.0  1.0
+ 1.0  1.0  1.0
+ 1.0  1.0  1.0
+
+```
+
+# See Also
+* [`flat_correct!`](@ref)
+"""
+flat_correct(frame::AbstractArray, flat_frame::AbstractArray; min_value = nothing, norm_value = nothing) = flat_correct!(deepcopy(frame), flat_frame, min_value = min_value, norm_value = norm_value)
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -77,34 +77,25 @@ subtract_overscan(frame::AbstractArray, idxs; dims = axes_min_length(idxs)) = su
 
 
 """
-    flat_correct!(frame::AbstractArray, flat_frame::AbstractArray; min_value = nothing, norm_value = nothing)
+    flat_correct!(frame::AbstractArray, flat_frame::AbstractArray; norm_value = mean(flat_frame))
 
 In place version of [`flat_correct`](@ref)
 """
-function flat_correct!(frame::AbstractArray, flat_frame::AbstractArray; min_value = nothing, norm_value = nothing)
+function flat_correct!(frame::AbstractArray, flat_frame::AbstractArray; norm_value = mean(flat_frame))
     size(frame) != size(flat_frame) && error("size of frame and flat frame are not same")
-    !(norm_value === nothing) && norm_value < 0 && error("norm_value must be greater than 0")
+    norm_value < 0 && error("norm_value must be greater than 0")
 
-    !(min_value === nothing) && @. flat_frame[flat_frame < min_value] = min_value
-
-    if !(norm_value === nothing)
-        flat_frame ./= norm_value
-    else
-        flat_frame ./= mean(flat_frame)
-    end
-
-    frame ./= flat_frame
+    frame ./= (flat_frame ./ norm_value)
     return frame
 end
 
 
 """
-    flat_correct(frame::AbstractArray, flat_frame::AbstractArray; min_value = nothing, norm_value = nothing)
+    flat_correct(frame::AbstractArray, flat_frame::AbstractArray; norm_value = mean(flat_frame))
 
 Correct `frame` for non-uniformity using the calibrated `flat_frame`.
 
-If `min_value` is provided, the `flat_frame` will be minimum-clipped to that value. By default,
-the `flat_frame` is normalized by its mean, but this can be changed by providing a custom `norm_value`.
+By default, the `flat_frame` is normalized by its mean, but this can be changed by providing a custom `norm_value`.
 
 !!! note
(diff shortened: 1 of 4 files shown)


src/methods.jl
```diff
@@ -77,3 +77,59 @@ julia> subtract_overscan(frame, (:, 4:5), dims = 2)
...
+
+    frame ./= flat_frame
+    return frame
+end
+
+
+"""
+    flat_correct(frame::AbstractArray, flat_frame::AbstractArray; min_value = nothing, norm_value = nothing)
+
+Correct the image for flat fielding.
```
mileslucas: This could be improved. I would write a short description like “Correct `frame` for non-uniformity using the calibrated `flat_frame`” then followed by a more description of the process and the keyword arguments. This would be a good place to note the behavior of dividing by zero (do we error, do we mask it, do we return infinity).

src/methods.jl
```diff
@@ -77,3 +77,59 @@ julia> subtract_overscan(frame, (:, 4:5), dims = 2)
...
+
+!!! note
```
mileslucas, line 95: I don’t feel like this needs to be an admonition

src/methods.jl
```diff
@@ -77,3 +77,59 @@ julia> subtract_overscan(frame, (:, 4:5), dims = 2)
...
+    The dimesions of `frame` and `flat_frame` should be identical.
+
+# Example
```
mileslucas: “Examples”

mileslucas (changes requested):

mileslucas (changes requested):
Code looks okay, Documentation could be a bit more thorough.

src/methods.jl
siddharthlal25: I had a similar admonition in `subtract_bias`, I will remove that as well.

src/methods.jl
```diff
@@ -77,3 +74,63 @@ julia> subtract_overscan(frame, (:, 4:5), dims = 2)
...
+
+    if !(norm_value === nothing)
+        flat_frame ./= norm_value
+    else
+        flat_frame ./= mean(flat_frame)
+    end
+
+    frame ./= flat_frame
+    return frame
+end
```
mileslucas: On second look, I tihnk you can avoid having to parse the keyword arguments by doing this-
```julia
function flat_correct!(frame, flat_frame; norm_value=mean(flat_frame), min_value=nothing)
    if min_value !== nothing
       @. flat_frame[flat_frame < min_value] = min_value
    end
    flat_normed = flat_frame ./ norm_value
    ...
end
```

src/methods.jl
```diff
@@ -77,3 +74,63 @@ julia> subtract_overscan(frame, (:, 4:5), dims = 2)
...
+
+
+"""
+    flat_correct(frame::AbstractArray, flat_frame::AbstractArray; min_value = nothing, norm_value = nothing)
+
+Correct `frame` for non-uniformity using the calibrated `flat_frame`.
+
+`min_value` is minimum value of `flat_frame`, an entry less than `min_value` is converted to `min_value` in `flat_frame`.
+`norm_value` is normalising constant for `flat_frame`, if set to `nothing` then `mean` of `flat_frame` is used for normalising.
```
mileslucas: The grammar here is a little sharp- it isn't that smooth to read, try

"If `min_value` is provided, the `flat_frame` will be minimum-clipped to that value. By default, the `flat_frame` is normalized by its mean, but this can be changed by providing a custom `norm_value`. "

src/methods.jl
```diff
@@ -77,3 +74,63 @@ julia> subtract_overscan(frame, (:, 4:5), dims = 2)
...
+
+!!! note
+    The function [`flat_correct`](@ref) has a division step of `frame` by `flat_frame`, if `flat_frame` has a zero entry then
+    the returned value is either `Inf` (dividend `frame` element is non-zero) or `NaN` (dividend `frame` element is zero).
```
mileslucas: A little awkward to reference the same docstring within the docstring-

"This function may introduce non-finite values if `flat_frame` contains values very close to `0` due to dividing by zero. The default behavior will return `Inf` if the `frame` value is non-zero, and `Nan` if the `frame` value is `0`."

mileslucas (changes requested):

src/methods.jl
siddharthlal25: Let me give this a try!

siddharthlal25: I don't think this is a good idea, let me explain:
suppose `flat = ones(3,3)` and let `min_value = 2`, we will not be setting up the value of `norm_value` explicitly so it is by default `norm_value = 1`, whereas as `min_value !== nothing` the `flat_frame` gets modified and the mean would now change to `2` and the `norm_value` should have been equal to `2` whereas it is `1`. Hence we have an incoming trouble.

siddharthlal25: Secondly, there is a fault with my version, it at times will act mutating for `flat_frame` as well (when some elements in the `flat_frame` is smaller than `min_value`), which I think is not something which a user would prefer. What are your thoughts on this? I was thinking of doing `_flat = deepcopy(flat_frame)` and then proceed in the usual way using `_flat`.

mileslucas:
Maybe it is just easier to remove the min_value functionality. I feel like
it makes more sense for the user to modify the flat frame beforehand
instead of adding this tiny bit to the function.

src/methods.jl
siddharthlal25: This would be always mutating for `flat_frame` (except for a few cases) as we are dividing the `flat_frame` by `norm_value` (`flat_frame ./= norm_value`), should we go for a `deepcopy`?

siddharthlal25: Resolved this issue by using fused broadcasts, i.e.
```
frame ./= (flat ./ norm_value)
```

src/methods.jl
```diff
@@ -77,3 +74,54 @@ julia> subtract_overscan(frame, (:, 4:5), dims = 2)
 * [`subtract_overscan!`](@ref)
 """
 subtract_overscan(frame::AbstractArray, idxs; dims = axes_min_length(idxs)) = subtract_overscan!(deepcopy(frame), idxs, dims = dims)
+
+
+"""
+    flat_correct!(frame::AbstractArray, flat_frame::AbstractArray; norm_value = mean(flat_frame))
+
+In place version of [`flat_correct`](@ref)
```
giordano:
```suggestion
In-place version of [`flat_correct`](@ref)
```

src/methods.jl
```diff
@@ -77,3 +74,54 @@ julia> subtract_overscan(frame, (:, 4:5), dims = 2)
...
+ 1.0  1.0  1.0
+ 1.0  1.0  1.0
+ 1.0  1.0  1.0
+
+```
+
+# See Also
+* [`flat_correct!`](@ref)
+"""
+flat_correct(frame::AbstractArray, flat_frame::AbstractArray; norm_value = mean(flat_frame)) = flat_correct!(deepcopy(frame), flat_frame, norm_value = norm_value)
```
giordano: Here I'd do:
```suggestion
flat_correct(frame::AbstractArray, flat_frame::AbstractArray; kwargs...) = flat_correct!(deepcopy(frame), flat_frame; kwargs...)
```
The advantage is that the list of keyword arguments is automatically inherited from `flat_correct!` and you don't need to manually keep them consistent

src/methods.jl
```diff
@@ -77,3 +74,54 @@ julia> subtract_overscan(frame, (:, 4:5), dims = 2)
...
+
+
+"""
+    flat_correct!(frame::AbstractArray, flat_frame::AbstractArray; norm_value = mean(flat_frame))
+
+In-place version of [`flat_correct`](@ref)
+"""
+function flat_correct!(frame::AbstractArray, flat_frame::AbstractArray; norm_value = mean(flat_frame))
+    size(frame) != size(flat_frame) && error("size of frame and flat frame are not same")
+    norm_value < 0 && error("norm_value must be greater than 0")
```
giordano: I think we can't accept zero, right? The error message looks correct (but I'd say "positive"), but the check isn't consistent with it

mileslucas: Yes, we should do `norm_value > 0 || error`

siddharthlal25: Just added an equal to sign as well!

src/methods.jl
```diff
@@ -77,3 +74,54 @@ julia> subtract_overscan(frame, (:, 4:5), dims = 2)
...
 subtract_overscan(frame::AbstractArray, idxs; dims = axes_min_length(idxs)) = subtract_overscan!(deepcopy(frame), idxs, dims = dims)
+
+
+"""
+    flat_correct!(frame::AbstractArray, flat_frame::AbstractArray; norm_value = mean(flat_frame))
+
+In-place version of [`flat_correct`](@ref)
+"""
+function flat_correct!(frame::AbstractArray, flat_frame::AbstractArray; norm_value = mean(flat_frame))
+    size(frame) != size(flat_frame) && error("size of frame and flat frame are not same")
```
mileslucas: So, I'm curious about needing this at all. This check is done automatically when we do the broadcasting, which also mentions the mismatched dimensions and size, so adding it here isn't really necessary other than saying `frame` and `flat frame` instead of `array A` and `array B`. Thoughts @giordano ?

giordano: Yes, when using broadcasting this is probably redundant. It's necessary when indexing directly, like what is done in `subtract_bias!`

siddharthlal25: Then let's switch to broadcasting in `subtract_bias!` as well, it maintains the uniformity in the code and also broadcast is GPU compatible. What would you suggest? @giordano @mileslucas

mileslucas: Sure, easy enough

mileslucas (approved):

giordano (approved):

PR #5, opened 2020-05-25
Added trim and trimview

siddharthlal25:
This PR adds `trim` and `trim!` to the package.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,8 +4,8 @@ authors = ["Siddharth Lal"]
 version = "0.1.0"
 
 [deps]
-Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 InvertedIndices = "41ab1584-1d38-5bbf-9106-f11c6c58b48f"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
 julia = "1"
--- a/Project.toml
+++ b/Project.toml
@@ -9,6 +9,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
 julia = "1"
+InvertedIndices = "1.0.0"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "0.1.0"
 
 [deps]
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+InvertedIndices = "41ab1584-1d38-5bbf-9106-f11c6c58b48f"
 
 [compat]
 julia = "1"
--- a/Project.toml
+++ b/Project.toml
@@ -4,12 +4,10 @@ authors = ["Siddharth Lal"]
 version = "0.1.0"
 
 [deps]
-InvertedIndices = "41ab1584-1d38-5bbf-9106-f11c6c58b48f"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
 julia = "1"
-InvertedIndices = "1.0.0"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -127,7 +127,7 @@ flat_correct(frame::AbstractArray, flat_frame::AbstractArray; kwargs...) = flat_
 
 Trims the frame to remove the region specified by idxs.
 
-This function trims the array in a manner such that final array should be a rectangle.
+This function trims the array in a manner such that final array should be rectangular.
 The indices follow standard Julia convention, so `(:, 45:60)` trims all columns from 45 to 60 and `(1:20, :)` trims all the rows from 1 to 20.
 
 # Examples
@@ -155,8 +155,7 @@ trim(frame::AbstractArray, idxs) = collect(trim!(frame, idxs))
 
 Trims the frame to remove the region specified by idxs.
 
-This function trims the array in a manner such that final array should be a rectangle.
-The indices follow standard Julia convention, so `(:, 45:60)` trims all columns from 45 to 60 and `(1:20, :)` trims all the rows from 1 to 20.
+This function is same as the [`trim`](@ref) function but returns a view of the frame.
 
 !!! note
     This function returns a view of the frame, so any modification to output
@@ -166,6 +165,7 @@ The indices follow standard Julia convention, so `(:, 45:60)` trims all columns 
 [`trim`](@ref)
 """
 function trim!(frame::AbstractArray, idxs)
+    # can switch to using `only` for Julia v1.4+
     ds = findall(x -> !isa(x, Colon), idxs)
     length(ds) == 1 || error("invalid trim indices $idxs")
 
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -167,14 +167,14 @@ This function is same as the [`trim`](@ref) function but returns a view of the f
 function trimview(frame::AbstractArray, idxs)
     # can switch to using `only` for Julia v1.4+
     ds = findall(x -> !isa(x, Colon), idxs)
-    length(ds) == 1 || error("invalid trim indices $idxs")
+    length(ds) == 1 || error("Invalid trim indices $idxs")
 
     d = ds[1]
+    full_idxs = axes(frame, d)
     # checking bounds error
-    idxs[d] ⊆ axes(frame, d) || error("out of bound trim indices $idxs")
+    idxs[d] ⊆ full_idxs || error("Trim indices $(idxs[d]) out of bounds for frame dimension $d $(full_idxs)")
 
     # finding the complement indices
-    full_idxs = deepcopy(axes(frame, d))
     complement_idxs = setdiff(full_idxs, idxs[d])
 
     return selectdim(frame, d, complement_idxs)
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -125,11 +125,11 @@ flat_correct(frame::AbstractArray, flat_frame::AbstractArray; kwargs...) = flat_
 """
     trim(frame::AbstractArray, idxs)
 
-    Trims the frame to remove the region specified by idxs.
+Trims the frame to remove the region specified by idxs.
 
-    This function trims the array in a manner such that final array should be a rectangle.
-    The input for idxs has to be of the form (:, 45:60) or (1:20, :) i.e. containing exactly one colon,
-    the former trims all columns from 45 to 60 whereas the later trims all the rows from 1 to 20.
+This function trims the array in a manner such that final array should be a rectangle.
+The input for idxs has to be of the form (:, 45:60) or (1:20, :) i.e. containing exactly one colon,
+the former trims all columns from 45 to 60 whereas the later trims all the rows from 1 to 20.
 
 !!! note
     This function returns a collected output of the trimmed view of frame, so any
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -128,12 +128,7 @@ flat_correct(frame::AbstractArray, flat_frame::AbstractArray; kwargs...) = flat_
 Trims the frame to remove the region specified by idxs.
 
 This function trims the array in a manner such that final array should be a rectangle.
-The input for idxs has to be of the form (:, 45:60) or (1:20, :) i.e. containing exactly one colon,
-the former trims all columns from 45 to 60 whereas the later trims all the rows from 1 to 20.
-
-!!! note
-    This function returns a collected output of the trimmed view of frame, so any
-    modifications to output array will not result in modification of frame.
+The indices follow standard Julia convention, so `(:, 45:60)` trims all columns from 45 to 60 and `(1:20, :)` trims all the rows from 1 to 20.
 
 # Examples
 ```jldoctest
@@ -152,12 +147,7 @@ julia> trim(frame, (:, 2:5))
 # See Also
 [`trim!`](@ref)
 """
-function trim(frame::AbstractArray, idxs)
-    ds = findall(x -> !isa(x, Colon), idxs)
-    length(ds) == 1 || error("invalid trim indices $idxs")
-
-    return selectdim(frame, ds[1], Not(idxs[ds[1]])) |> Array
-end
+trim(frame::AbstractArray, idxs) = collect(trim!(frame, idxs))
 
 
 """
@@ -166,8 +156,7 @@ end
 Trims the frame to remove the region specified by idxs.
 
 This function trims the array in a manner such that final array should be a rectangle.
-The input for idxs has to be of the form (:, 45:60) or (1:20, :) i.e. containing exactly one colon,
-the former trims all columns from 45 to 60 whereas the later trims all the rows from 1 to 20.
+The indices follow standard Julia convention, so `(:, 45:60)` trims all columns from 45 to 60 and `(1:20, :)` trims all the rows from 1 to 20.
 
 !!! note
     This function returns a view of the frame, so any modification to output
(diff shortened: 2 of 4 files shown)


Project.toml
```diff
@@ -4,6 +4,7 @@ authors = ["Siddharth Lal"]
 version = "0.1.0"
 
 [deps]
+InvertedIndices = "41ab1584-1d38-5bbf-9106-f11c6c58b48f"
```
mileslucas: please add a compat entry

src/methods.jl
```diff
@@ -120,3 +120,65 @@ julia> flat_correct(frame, flat)
...
+
+
+"""
+    trim(frame::AbstractArray, idxs)
+
+Trims the frame to remove the region specified by idxs.
+
+This function trims the array in a manner such that final array should be a rectangle.
+The input for idxs has to be of the form (:, 45:60) or (1:20, :) i.e. containing exactly one colon,
+the former trims all columns from 45 to 60 whereas the later trims all the rows from 1 to 20.
```
```suggestion
This function trims the array in a manner such that final array should be a rectangular.
The indices follow standard Julia convention, so `(:, 45:60)` trims all columns from 45 to 60 and `(1:20, :)` trims all the rows from 1 to 20.
```

src/methods.jl
```diff
@@ -120,3 +120,65 @@ julia> flat_correct(frame, flat)
...
+
+!!! note
+    This function returns a collected output of the trimmed view of frame, so any
+    modifications to output array will not result in modification of frame.
```
mileslucas: this seems unnecessary. This is "normal" behavior.

src/methods.jl
```diff
@@ -120,3 +120,65 @@ julia> flat_correct(frame, flat)
...
+
+# See Also
+[`trim!`](@ref)
+"""
+function trim(frame::AbstractArray, idxs)
+    ds = findall(x -> !isa(x, Colon), idxs)
+    length(ds) == 1 || error("invalid trim indices $idxs")
+
+    return selectdim(frame, ds[1], Not(idxs[ds[1]])) |> Array
+end
```
```suggestion
trim(frame, idxs) = collect(trim!(frame, idxs))
```

src/methods.jl
```diff
@@ -120,3 +120,65 @@ julia> flat_correct(frame, flat)
...
+    array will result in modification of frame.
+
+# See Also
+[`trim`](@ref)
+"""
+function trim!(frame::AbstractArray, idxs)
+    ds = findall(x -> !isa(x, Colon), idxs)
+    length(ds) == 1 || error("invalid trim indices $idxs")
+
+    return selectdim(frame, ds[1], Not(idxs[ds[1]]))
```
mileslucas: if we add `Compat.jl` this can be reduced to 
```julia
@compat d = only(ds)
return selectdim(frame, d, Not(idxs[d]))
```
because `only` will fail if `ds` doesn't have length 1. I'm not sure it's necessarily worth it, though.

test/methods.jl
```diff
@@ -45,6 +48,16 @@ end
     @test_throws ErrorException flat_correct(ones(5, 5), ones(5, 5), norm_value = -2)
 end
 
+@testset "trim" begin
+    @test trim(ones(5, 5), (:, 4:5)) == trim!(ones(5, 5), (:, 4:5))
+    @test trim(ones(5, 5), (4:5, :)) == trim!(ones(5, 5), (4:5, :))
+
```
mileslucas, line 58: would be nice to test that `trim!` results in a view

mileslucas (changes requested):

src/methods.jl
siddharthlal25: I will make a final commit where I will remove all `AbstractArray` keywords.

src/methods.jl
siddharthlal25: I don't think it would be of much worth, let's try and keep our dependencies as less as possible.

mileslucas: Okay. I think adding a comment along the lines of "# Julia v1.4+ can switch to using `only`" could be helpful.

src/methods.jl
```diff
@@ -120,3 +120,54 @@ julia> flat_correct(frame, flat)
...
 """
 flat_correct(frame::AbstractArray, flat_frame::AbstractArray; kwargs...) = flat_correct!(deepcopy(frame), flat_frame; kwargs...)
+
+
+"""
+    trim(frame::AbstractArray, idxs)
+
+Trims the frame to remove the region specified by idxs.
+
+This function trims the array in a manner such that final array should be a rectangle.
```
mileslucas: I specifically changed to "rectangular" because "rectangle" is ambiguous. "the final array should be a rectangle"- does that imply the shape or that the literal array has become a rectangle (ie some single valued array).

src/methods.jl
```diff
@@ -120,3 +120,54 @@ julia> flat_correct(frame, flat)
...
+trim(frame::AbstractArray, idxs) = collect(trim!(frame, idxs))
+
+
+"""
+    trim!(frame::AbstractArray, idxs)
+
+Trims the frame to remove the region specified by idxs.
+
+This function trims the array in a manner such that final array should be a rectangle.
+The indices follow standard Julia convention, so `(:, 45:60)` trims all columns from 45 to 60 and `(1:20, :)` trims all the rows from 1 to 20.
```
mileslucas: Can probably simplify to something along the lines of "the same as trim but returns a view"

src/methods.jl

src/methods.jl
siddharthlal25: Sorry, I missed it, I will change it as soon as I wake up!

src/methods.jl

src/methods.jl
```diff
@@ -120,3 +120,54 @@ julia> flat_correct(frame, flat)
...
+```
+
+# See Also
+[`trim!`](@ref)
+"""
+trim(frame::AbstractArray, idxs) = collect(trim!(frame, idxs))
+
+
+"""
+    trim!(frame::AbstractArray, idxs)
```
giordano: I'm confused by the use of `!` here: this function doesn't really modify anything, does it?

src/methods.jl
```diff
@@ -120,3 +120,54 @@ julia> flat_correct(frame, flat)
...
+ 1.0
+ 1.0
+ 1.0
+
+```
+
+# See Also
+[`trim!`](@ref)
+"""
+trim(frame::AbstractArray, idxs) = collect(trim!(frame, idxs))
```
giordano: `copy` or `deepcopy` are probably more idiomatic than `collect`?

src/methods.jl
siddharthlal25: This function is not mutating in the subroutine, but when the returned array is modified the parent array is also modified, that's why I used `!`. What's your thoughts on the use of `!`?

src/methods.jl
siddharthlal25: So, should we switch to something like 
```
trim(...) = deepcopy(trim!(...))
```
Right?

src/methods.jl
giordano: Let's first make it clear that the `!` is purely a convention, nothing fundamental.  [`view`](https://docs.julialang.org/en/v1/base/arrays/#Base.view) behaves exactly like your `trim!` and doesn't have a bang in its name.

To be honest, I fail to see what these trimming functions do that basic indexing in Julia already doesn't.

test/methods.jl
```diff
@@ -45,6 +48,20 @@ end
     @test_throws ErrorException flat_correct(ones(5, 5), ones(5, 5), norm_value = -2)
 end
 
+@testset "trim" begin
+    @test trim(ones(5, 5), (:, 4:5)) == trim!(ones(5, 5), (:, 4:5))
+    @test trim(ones(5, 5), (4:5, :)) == trim!(ones(5, 5), (4:5, :))
```
giordano: Note that these tests are not particularly useful: you're basically testing that the trimmed arrays on both sides have the same shape, not the same content, since they're all ones anyway.  You should be using a different array to test with, for example something like `reshape(1:25, 5, 5)` or `rand(5, 5)`

src/methods.jl
giordano: > To be honest, I fail to see what these trimming functions do that basic indexing in Julia already doesn't.

Sorry, yeah, the difference is that `trim` indexes the complementary indices

test/methods.jl
siddharthlal25: Sure! I will modify the tests.

src/methods.jl
siddharthlal25: I tried both these versions, here are the results (I defined new function `trim1` with `deepcopy`)
```
trim1(frame::AbstractArray, idxs) = deepcopy(trim!(frame, idxs))
```

The results were:
```
julia> x = rand(4096, 4096);

julia> @benchmark trim(x, (:, 3000:4096))
BenchmarkTools.Trial: 
  memory estimate:  94.50 MiB
  allocs estimate:  22993
  --------------
  minimum time:     77.609 ms (0.00% GC)
  median time:      78.180 ms (0.39% GC)
  mean time:        83.516 ms (6.98% GC)
  maximum time:     147.196 ms (47.21% GC)
  --------------
  samples:          60
  evals/sample:     1

julia> @benchmark trim1(x, (:, 3000:4096))
BenchmarkTools.Trial: 
  memory estimate:  128.81 MiB
  allocs estimate:  23002
  --------------
  minimum time:     92.075 ms (0.00% GC)
  median time:      103.485 ms (0.31% GC)
  mean time:        102.107 ms (7.45% GC)
  maximum time:     166.797 ms (44.01% GC)
  --------------
  samples:          50
  evals/sample:     1
```
Memory and time utilised in `trim1` with `deepcopy` is more as compared to `trim`, so I guess it's better to use `trim`. What causes this kind of difference in utilised memory and time?

giordano: Did you try `copy`?

siddharthlal25: `copy` is faster!
```
julia> trim2(frame::AbstractArray, idxs) = copy(trim!(frame, idxs));

julia> x = randn(4096, 4096);

julia> @benchmark trim2(x, (:, 3000:4096))
BenchmarkTools.Trial: 
  memory estimate:  94.50 MiB
  allocs estimate:  22995
  --------------
  minimum time:     65.307 ms (0.00% GC)
  median time:      67.059 ms (0.40% GC)
  mean time:        71.168 ms (7.81% GC)
  maximum time:     136.852 ms (51.54% GC)
  --------------
  samples:          71
  evals/sample:     1
```

src/methods.jl
mileslucas: Yeah I thought about this for a bit- I would love for there to be a version that returns a view because it will be faster if it’s a piece within a pipeline. However I don’t think view should be the default, since an average wandering astronomer will inevitably overwrite their data on accident. The bang method was all I could think at the time to differentiate the two, how else could we differentiate the two?

giordano: Yes, I honestly can't offer a better alternative which isn't too verbose.  However, note that while being faster than the `trim` function, this `trim!` allocates quite a bit:
```julia
julia> A = rand(4096, 4096);

julia> @benchmark trim!($A, $((:, 3000:4096)))
BenchmarkTools.Trial: 
  memory estimate:  804.89 KiB
  allocs estimate:  22991
  --------------
  minimum time:     1.184 ms (0.00% GC)
  median time:      1.430 ms (0.00% GC)
  mean time:        1.426 ms (5.04% GC)
  maximum time:     14.958 ms (90.78% GC)
  --------------
  samples:          3501
  evals/sample:     1

julia> @benchmark view($A, $(:), $(1:2999))
BenchmarkTools.Trial: 
  memory estimate:  64 bytes
  allocs estimate:  1
  --------------
  minimum time:     7.858 ns (0.00% GC)
  median time:      25.282 ns (0.00% GC)
  mean time:        27.180 ns (19.60% GC)
  maximum time:     13.698 μs (99.88% GC)
  --------------
  samples:          10000
  evals/sample:     999
```

siddharthlal25: I think these allocations are because of `Not` of `InvertedIndices` and the inner design of `InvertedIndices`. Should we switch to some if-else conditions for `trim!`? this would lead to fewer allocations but the code loses its beauty of being short and succinct

mileslucas: I’m surprised about the number of allocations, the underlying code that parses the indices shouldn’t be allocating on the surface. 


For the naming, let’s use `trim` and `trimview`. This is the notation used by a few of the JuliaImages functions (warpedview)

src/methods.jl
siddharthlal25: I missed interpolating the variables, let me do it with interpolation!

siddharthlal25: `trim` - with collect
`trim1` - with deepcopy
`trim2` - with copy
```
julia> x = rand(4096, 4096);

julia> @benchmark trim($x, $((:, 3000:4096)))
BenchmarkTools.Trial: 
  memory estimate:  94.50 MiB
  allocs estimate:  22993
  --------------
  minimum time:     43.108 ms (0.00% GC)
  median time:      44.161 ms (0.59% GC)
  mean time:        46.359 ms (6.48% GC)
  maximum time:     91.735 ms (52.76% GC)
  --------------
  samples:          108
  evals/sample:     1

julia> @benchmark trim1($x, $((:, 3000:4096)))
BenchmarkTools.Trial: 
  memory estimate:  128.81 MiB
  allocs estimate:  23002
  --------------
  minimum time:     52.172 ms (0.00% GC)
  median time:      53.200 ms (0.36% GC)
  mean time:        56.204 ms (6.78% GC)
  maximum time:     102.094 ms (48.88% GC)
  --------------
  samples:          89
  evals/sample:     1

julia> @benchmark trim2($x, $((:, 3000:4096)))
BenchmarkTools.Trial: 
  memory estimate:  94.50 MiB
  allocs estimate:  22995
  --------------
  minimum time:     36.425 ms (0.00% GC)
  median time:      37.791 ms (0.52% GC)
  mean time:        40.183 ms (7.50% GC)
  maximum time:     85.394 ms (57.44% GC)
  --------------
  samples:          125
  evals/sample:     1
```

src/methods.jl
siddharthlal25: So, should we use `InvertedIndices` with more allocations and succinct code or switch to if-else conditions and use `@view`?

siddharthlal25: Removed using `InvertedIndices` and switched to using `setdiff` for generating complement indices, this reduced allocations by a factor of 4.

src/methods.jl
siddharthlal25: Used `copy` as it was the fastest amongst all!

src/methods.jl
```diff
@@ -120,3 +120,62 @@ julia> flat_correct(frame, flat)
...
+[`trim`](@ref)
+"""
+function trimview(frame::AbstractArray, idxs)
+    # can switch to using `only` for Julia v1.4+
+    ds = findall(x -> !isa(x, Colon), idxs)
+    length(ds) == 1 || error("invalid trim indices $idxs")
+
+    d = ds[1]
+    # checking bounds error
+    idxs[d] ⊆ axes(frame, d) || error("out of bound trim indices $idxs")
```
mileslucas: take a look at default indexing errors-
```
ERROR: BoundsError: attempt to access 10×10 Array{Float64,2} at index [Base.Slice(Base.OneTo(10)), 3:11]
```
it shows both the expectation (implicitly here, but 10x10 array implies the correct indices are 1 to 10) and the given `3:11`. 

I would recommend showing an error that includes the expectation- something like

```julia
error("Trim indices $(idxs[d]) out of bounds for frame dimension $d $(axes(frame, d))")
```

src/methods.jl
```diff
@@ -120,3 +120,62 @@ julia> flat_correct(frame, flat)
...
+
+    # finding the complement indices
+    full_idxs = deepcopy(axes(frame, d))
```
mileslucas: unnecessary `deepcopy`, also consider putting this before your bounds checking so you don't have multiple calls to `axes(frame, d)`

mileslucas (changes requested):

src/methods.jl
siddharthlal25: Didn't change it after experimenting, will correct it asap.

mileslucas (approved):

giordano (approved):

siddharthlal25:
I had missed the tests for type-stability, will add it with the PR for `crop` and `cropview`.

PR #7, opened 2020-05-27
Added crop and cropview

siddharthlal25:
This PR adds `crop` and `cropview` for this package.

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -195,7 +195,8 @@ end
 
 Crops `frame` to the size specified by `shape` anchored by the frame center.
 
-This will remove rows/cols of the `frame` equally on each side. When there is an uneven difference in sizes (e.g. size 9 -> 6 can't be removed equally) the default is to increase the output size (e.g. 6 -> 7) so there is equal removal on each side. To disable this, set `force_equal=false`, which will remove the extra slice from the end of the axis.
+This will remove rows/cols of the `frame` equally on each side. When there is an uneven difference in sizes (e.g. size 9 -> 6 can't be removed equally) the default is to 
+increase the output size (e.g. 6 -> 7) so there is equal removal on each side. To disable this, set `force_equal=false`, which will remove the extra slice from the end of the axis.
 
 # Examples
 ```jldoctest
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -179,3 +179,63 @@ function trimview(frame::AbstractArray, idxs)
 
     return selectdim(frame, d, complement_idxs)
 end
+
+
+"""
+    crop(frame::AbstractArray, shape; force_odd = true)
+
+Crops the image to the size specified asumming center as the anchor.
+
+# See Also
+* [`cropview`](@ref)
+"""
+crop(frame::AbstractArray, shape; kwargs...) = copy(cropview(frame, shape; kwargs...))
+
+
+"""
+    cropview(frame::AbstractArray, shape; force_odd = true)
+
+Crops the image to the size specified asumming center as the anchor.
+
+!!! note
+    This function returns a view of the frame, so any modification to output
+    array will result in modification of frame.
+
+# See Also
+* [`crop`](@ref)
+"""
+function cropview(frame::AbstractArray, shape; force_odd = true)
+    # testing error
+
+
+    # generating the row range in cropped view
+    rows = size(frame, 1)
+    if (rows - shape[1])%2 == 0
+        row_range = 1 + (rows - shape[1]) ÷ 2 : rows - (rows - shape[1]) ÷ 2
+    else
+        if force_odd
+            @warn "Diffence between row dimension of cropped view and frame is odd. Crop row dimension increased by one."
+            mod_x = shape[1] + 1
+            row_range = 1 + (rows - mod_x) ÷ 2 : rows - (rows - mod_x) ÷ 2
+        else
+            row_range = 1 + (rows - shape[1] - 1) ÷ 2 : rows - (rows - shape[1] + 1) ÷ 2
+        end
+    end
+
+    # generating the col range in cropped view
+    cols = size(frame, 2)
+    if (cols - shape[2])%2 == 0
+        col_range = 1 + (cols - shape[2]) ÷ 2 : cols - (cols - shape[2]) ÷ 2
+    else
+        if force_odd
+            @warn "Diffence between column dimension of cropped view and frame is odd. Crop column dimension increased by one."
+            mod_y = shape[2] + 1
+            col_range = 1 + (cols - mod_y) ÷ 2 : cols - (cols - mod_y) ÷ 2
+        else
+            col_range = 1 + (cols - shape[2] - 1) ÷ 2 : cols - (cols - shape[2] + 1) ÷ 2
+        end
+    end
+
+    # returning the view
+    return @view frame[row_range, col_range]
+end
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -145,7 +145,7 @@ julia> trim(frame, (:, 2:5))
 ```
 
 # See Also
-[`trimview`](@ref)
+* [`trimview`](@ref)
 """
 trim(frame::AbstractArray, idxs) = copy(trimview(frame, idxs))
 
@@ -162,7 +162,7 @@ This function is same as the [`trim`](@ref) function but returns a view of the f
     array will result in modification of frame.
 
 # See Also
-[`trim`](@ref)
+* [`trim`](@ref)
 """
 function trimview(frame::AbstractArray, idxs)
     # can switch to using `only` for Julia v1.4+
@@ -236,8 +236,7 @@ function cropview(frame::AbstractArray, shape; force_equal = true)
             push!(range, 1 + shape_diff[i] ÷ 2 : strip - shape_diff[i] ÷ 2)
         else
             if force_equal
-                i == 1 && @warn "Diffence between row dimension of cropped view and frame is odd. Crop row dimension increased by one."
-                i == 2 && @warn "Diffence between column dimension of cropped view and frame is odd. Crop column dimension increased by one."
+                @warn "Shape was increased to keep equal trim size"
                 push!(range, 1 + (shape_diff[i] - 1) ÷ 2 : strip - (shape_diff[i] - 1) ÷ 2)
             else
                 push!(range, 1 + (shape_diff[i] - 1) ÷ 2 : strip - (shape_diff[i] + 1) ÷ 2)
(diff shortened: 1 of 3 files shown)


siddharthlal25:
@mileslucas @giordano suggest me some better names for variables used inside the function.

src/methods.jl
```diff
@@ -179,3 +179,63 @@ function trimview(frame::AbstractArray, idxs)
...
+        col_range = 1 + (cols - shape[2]) ÷ 2 : cols - (cols - shape[2]) ÷ 2
+    else
+        if force_odd
+            @warn "Diffence between column dimension of cropped view and frame is odd. Crop column dimension increased by one."
+            mod_y = shape[2] + 1
+            col_range = 1 + (cols - mod_y) ÷ 2 : cols - (cols - mod_y) ÷ 2
+        else
+            col_range = 1 + (cols - shape[2] - 1) ÷ 2 : cols - (cols - shape[2] + 1) ÷ 2
+        end
+    end
```
mileslucas: I think we can organize this better- keep in mind that `shape` and `size(frame)` are both tuples of integers and remember the "Don't Repeat Yourself" principle. I see the same few things repeated multiple times, sometimes in the same line!

Why don't you start with this-
```julia
shape_diff = size(frame) .- shape
```
and see how you can simplify this code. Also keep in mind functions like `isodd` and `iseven`.

src/methods.jl
```diff
@@ -179,3 +179,63 @@ function trimview(frame::AbstractArray, idxs)
 
     return selectdim(frame, d, complement_idxs)
 end
+
+
+"""
+    crop(frame::AbstractArray, shape; force_odd = true)
+
+Crops the image to the size specified asumming center as the anchor.
```
mileslucas: take a look at the project notes for more ideas for documentation.

mileslucas (changes requested):

src/methods.jl
mileslucas: immediately you can add a test 
```julia
any(s -> s < 0, shape_diff) && error("crop size ($shape) can't be larger than frame size ($(size(frame)))")
```
, in one line!

siddharthlal25: Added the test and reduced the size of function by keeping in mind the DRY principle!

src/methods.jl
siddharthlal25: I have added a few docstrings!

src/methods.jl
```diff
@@ -179,3 +179,72 @@ function trimview(frame::AbstractArray, idxs)
...
+    # calculating range of cropped frame
+    range = []
+    for i = 1:2
+        strip = size(frame, i)
+        if iseven(shape_diff[i])
+            push!(range, 1 + shape_diff[i] ÷ 2 : strip - shape_diff[i] ÷ 2)
+        else
+            if force_equal
+                i == 1 && @warn "Diffence between row dimension of cropped view and frame is odd. Crop row dimension increased by one."
+                i == 2 && @warn "Diffence between column dimension of cropped view and frame is odd. Crop column dimension increased by one."
```
siddharthlal25: Any smart way to handle these warnings? @giordano @mileslucas

src/methods.jl
```diff
@@ -179,3 +179,72 @@ function trimview(frame::AbstractArray, idxs)
...
 end
+
+
+"""
+    crop(frame::AbstractArray, shape; force_equal = true)
+
+Crops the image to the shape specified asumming center as the anchor.
+
+Shape speficies the size of the cropped array, force_equal when set to true forces the cropped rows and columns to be equal on both sides
+otherwise the cropped frame contains one extra column on left and one extra row on top if necessary.
```
siddharthlal25: Thoughts about the documentation? @giordano @mileslucas

src/methods.jl
mileslucas: Why do you need to be so granular? How about if any dimension is changed you put one single succinct warning, this warning might not even fit on one line

mileslucas: “Shape was increased to keep equal trim size”

src/methods.jl
```diff
@@ -179,3 +179,71 @@ function trimview(frame::AbstractArray, idxs)
...
+"""
+function cropview(frame::AbstractArray, shape; force_equal = true)
+    shape_diff = size(frame) .- shape
+
+    # testing error
+    any(s -> s < 0, shape_diff) && error("crop size ($shape) can't be larger than frame size ($(size(frame)))")
+    any(s -> s < 0, shape) && error("crop size ($shape) cant't be negative")
+
+    # calculating range of cropped frame
+    range = []
```
giordano: I'd initialise with
```suggestion
    range = Vector{Int}(undef, 2)
```
and use assignment below, rather than `push!`ing

siddharthlal25: This should probably be: `range = Vector{Vector{Int}}(undef, 2)` because we are storing ranges in this.

src/methods.jl
```diff
@@ -179,3 +179,71 @@ function trimview(frame::AbstractArray, idxs)
...
+    for i = 1:2
+        strip = size(frame, i)
+        if iseven(shape_diff[i])
+            push!(range, 1 + shape_diff[i] ÷ 2 : strip - shape_diff[i] ÷ 2)
+        else
+            if force_equal
+                @warn "Shape was increased to keep equal trim size"
+                push!(range, 1 + (shape_diff[i] - 1) ÷ 2 : strip - (shape_diff[i] - 1) ÷ 2)
+            else
+                push!(range, 1 + (shape_diff[i] - 1) ÷ 2 : strip - (shape_diff[i] + 1) ÷ 2)
```
giordano: Oh, right, I didn't notice this is a range.  Pay attention to precedence of operators here, you may not be doing what you expect here.  Use parentheses as appropriate

src/methods.jl
giordano: Then
```suggestion
    range = Vector{UnitRange{Int}}(undef, 2)
```

src/methods.jl
siddharthlal25: I wanted to divide first, so I think this would work fine! Still, I will put the parentheses.

giordano: My worry is in particular about `:`, it has a weird precedence.  When I have ranges more complex than `a:b` I prefer to wrap both sides of `:` with parentheses: `(a + b * c) : (d ^ e / f)`

src/methods.jl
siddharthlal25: Switched to `maps` in place of `for` loop!

src/methods.jl
siddharthlal25: Added the parentheses!

src/methods.jl
```diff
@@ -145,7 +145,7 @@ julia> trim(frame, (:, 2:5))
 ```
 
 # See Also
-[`trimview`](@ref)
+* [`trimview`](@ref)
```
mileslucas: I actually think a comma separated list is easier to read and less distracting than a bulleted list.

src/methods.jl
```diff
@@ -179,3 +179,66 @@ function trimview(frame::AbstractArray, idxs)
 
     return selectdim(frame, d, complement_idxs)
 end
+
+
+"""
+    crop(frame::AbstractArray, shape; force_equal = true)
+
+Crops the image to the shape specified asumming center as the anchor.
```
```suggestion
Crops `frame` to the size specified by `shape` anchored by the frame center.
```

src/methods.jl
mileslucas:
```suggestion
This will remove rows/cols of the `frame` equally on each side. When there is an uneven difference in sizes (e.g. size 9 -> 6 can't be removed equally) the default is to increase the output size (e.g. 6 -> 7) so there is equal removal on each side. To disable this, set `force_equal=false`, which will remove the extra slice from the end of the axis.
```

src/methods.jl
```diff
@@ -179,3 +179,66 @@ function trimview(frame::AbstractArray, idxs)
...
+# Examples
+```jldoctest
+julia> frame = reshape(1:25, (5, 5));
+
+julia> crop(frame, (3, 3))
+3×3 Array{Int64,2}:
+ 7  12  17
+ 8  13  18
+ 9  14  19
+
```
mileslucas, line 210: you should show an example with `force_equal=false` for the docstring

src/methods.jl
```diff
@@ -179,3 +179,66 @@ function trimview(frame::AbstractArray, idxs)
...
+                diff = s2 isa Colon ? 0 : s1 - s2
+                lower = iseven(diff) ? (1 + (diff ÷ 2)) : (1 + ((diff - 1) ÷ 2))
+                upper = if isodd(diff) && force_equal
+                            @warn "dimension $d changed from $s2 to $(s2 + 1)"
+                            (s1 - ((diff - 1) ÷ 2))
+                        else
+                            iseven(diff) ? (s1 - (diff ÷ 2)) : (s1 - ((diff + 1) ÷ 2))
+                        end
+                lower:upper
+            end
```
```suggestion
    idxs = map(enumerate(size(frame)), shape) do (d, s1), s2
                diff = s2 isa Colon ? 0 : s1 - s2
                lower = iseven(diff) ? diff ÷ 2 : (diff - 1) ÷ 2
                upper = if isodd(diff) && force_equal
                            @warn "dimension $d changed from $s2 to $(s2 + 1)"
                            (diff - 1) ÷ 2
                        elseif isodd(diff)
                            (diff + 1) ÷ 2
                        else
                            diff ÷ 2
                        end
                1 + lower:s1 - upper
            end
```

just a little easier to parse.

mileslucas (changes requested):

src/methods.jl
```diff
@@ -179,3 +188,74 @@ function trimview(frame::AbstractArray, idxs)
...
+                lower = iseven(diff) ? (diff ÷ 2) : ((diff - 1) ÷ 2)
+                upper = if isodd(diff) && force_equal
+                            @warn "dimension $d changed from $s2 to $(s2 + 1)"
+                            (diff - 1) ÷ 2
+                        elseif isodd(diff)
+                            (diff + 1) ÷ 2
+                        else
+                            diff ÷ 2
+                        end
+                (1 + lower):(s1 - upper)
```
mileslucas: You don’t need to wrap these in parantheses IIRC, the : operator has extremely low precedence.

mileslucas (approved):

src/methods.jl
siddharthlal25: Okay!

siddharthlal25: I think let's keep the brackets, @giordano also suggested to keep them because of some strange behaviours of precedence. Thoughts?

giordano: That's more a personal taste, I don't mind too much as long as the result is correct

siddharthlal25: Currently, it works perfectly fine without brackets. So, let's keep them away! 😄

giordano (approved):

PR #10, opened 2020-05-28
Added support for FITS based indices in `trim` and `subtract_overscan`

### LINKED ISSUE #8: FITS style indexing for `trim` and `subtract_overscan`
The functions `trim`, `trimview`, `subtract_overscan` and `subtract_overscan!` all can use normal Julian indices (current implementation) but we can also support FITS style indexing. 

I would suggest creating a function the takes FITS-style indices and translates them into Julian indices
```julia
fits_indices(idxs::String)
```
Then you can expand the other function like this
```julia
f(frame, idxs::String) = f(frame, fits_indices(idxs))
```

Now, here's a fun trick- since each of these pairs of functions calls another, you only have to specialize one of them! (here's an example-

```julia
# only need one here, unspecialized
subtract_overscan(frame, idxs) = subtract_overscan!(copy(frame), idxs)
# specialize here
subtract_overscan!(frame, idxs::String) = subtract_overscan!(frame, fits_indices(idxs))
function subtract_ovserscan!(frame, idxs::Union{<:Tuple, <:AbstractVector})
    # this code already exists
end
```

siddharthlal25:
Fixes #8

### DIFF
--- a/src/CCDReduction.jl
+++ b/src/CCDReduction.jl
@@ -11,7 +11,8 @@ export subtract_bias,
        trim,
        trimview,
        crop,
-       cropview
+       cropview,
+       fits_indices
 
 include("methods.jl")
 
--- a/src/CCDReduction.jl
+++ b/src/CCDReduction.jl
@@ -11,8 +11,7 @@ export subtract_bias,
        trim,
        trimview,
        crop,
-       cropview,
-       fits_indices
+       cropview
 
 include("methods.jl")
 
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -9,25 +9,20 @@ Converts FITS based indices to Julian indices.
 
 # Examples
 ```jldoctest
-julia> fits_indices("[1024:2048, :]")
+julia> fits_indices("[1024:2048, 1:2048]")
 2-element Array{Any,1}:
- Colon()
+ 1:2048
  1024:2048
 
 ```
 """
 function fits_indices(string::String)
-    str = replace(replace(replace(string, "[" => ""), "]" => ""), " " => "")
+    str = replace(string, r"[\[\]\s]" => "")
     tokens = split(str, ',')
 
     idxs = map(tokens) do token
-        t = split(token, ':')
-        if all(i -> i == "", t)
-            Colon()
-        else
-            low, high = parse.(Int, t)
-            low : high
-        end
+        t = split(token, ':', keepempty=false)
+        length(t) == 0 ? Colon() : parse(Int, t[1]):parse(Int, t[2])
     end
     return reverse(idxs)
 end
@@ -223,7 +218,7 @@ This function is same as the [`trim`](@ref) function but returns a view of the f
 """
 function trimview(frame::AbstractArray, idxs)
     # this adds the support for input indices of the form (1:size(frame, 1), ...) or (..., 1:size(frame, 2))
-    # It converts 1:size(frame, 1) to : 1and then the same subroutine follows.
+    # It converts 1:size(frame, 1) to : and then the same subroutine follows.
     processed_idxs = map(size(frame), idxs) do s1, s2
                             if s2 == Colon()
                                 Colon()
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -1,5 +1,6 @@
 # helper function
 axes_min_length(idxs) = argmin([a isa Colon ? Inf : length(a) for a in idxs])
+axes_min_length(idxs::String) = axes_min_length(fits_indices(idxs))
 
 """
     fits_indices(string::String)
@@ -83,6 +84,8 @@ function subtract_overscan!(frame::AbstractArray, idxs; dims = axes_min_length(i
     frame .-= overscan_value
     return frame
 end
+
+subtract_overscan!(frame::AbstractArray, idxs::String; kwargs...) = subtract_overscan!(frame, fits_indices(idxs); kwargs...)
 
 
 """
@@ -218,6 +221,8 @@ function trimview(frame::AbstractArray, idxs)
 
     return selectdim(frame, d, complement_idxs)
 end
+
+trimview(frame::AbstractArray, idxs::String) = trimview(frame, fits_indices(idxs))
 
 
 """
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -191,7 +191,7 @@ julia> trim(frame, (:, 2:5))
  1.0
  1.0
 
-julia> trim(frame, "[2:5, :]")
+julia> trim(frame, "[2:5, 1:5]")
 5×1 Array{Float64,2}:
  1.0
  1.0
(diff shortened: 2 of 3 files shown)


src/CCDReduction.jl
```diff
@@ -11,7 +11,8 @@ export subtract_bias,
        trim,
        trimview,
        crop,
-       cropview
+       cropview,
+       fits_indices
```
mileslucas: I wouldn't export this

src/methods.jl
```diff
@@ -1,5 +1,36 @@
...
+```jldoctest
+julia> fits_indices("[1024:2048, :]")
+2-element Array{Any,1}:
+ Colon()
+ 1024:2048
+
+```
+"""
+function fits_indices(string::String)
+    str = replace(replace(replace(string, "[" => ""), "]" => ""), " " => "")
```
```suggestion
    str = replace(string, r"[\[\]\s]" => "")
```

src/methods.jl
```diff
@@ -1,5 +1,36 @@
...
+    tokens = split(str, ',')
+
+    idxs = map(tokens) do token
+        t = split(token, ':')
+        if all(i -> i == "", t)
+            Colon()
+        else
+            low, high = parse.(Int, t)
+            low : high
+        end
```
mileslucas: technically `"[1024:2048, :]"` is not a valid FITS index, it has to be `"[1024:2048, 1:2048]"`. I would change this in the example, but I think it's a good idea to keep this code.
```suggestion
        t = split(token, ':', keepempty=false)
        length(t) == 0 ? Colon() : parse(Int, t[1]):parse(Int, t[2])
```

src/methods.jl
```diff
@@ -1,5 +1,36 @@
 # helper function
 axes_min_length(idxs) = argmin([a isa Colon ? Inf : length(a) for a in idxs])
+axes_min_length(idxs::String) = axes_min_length(fits_indices(idxs))
```
mileslucas: why do you need this?

test/methods.jl
```diff
@@ -111,4 +119,9 @@ end
     @test axes_min_length((1:2, :)) == 1
     @test axes_min_length((1:2, 5:12)) == 1
     @test axes_min_length((10:100, 1:5)) == 2
+
+    # testing fits_indices
+    @test fits_indices("[1024:2048, 200:300]") == [200:300, 1024:2048]
+    @test fits_indices("[:, 200:300]") == [200:300, :]
+    @test fits_indices("[1024:2048, :]") == [:, 1024:2048]
```
mileslucas, line 129: again, these technically aren't valid FITS indices- but keep these tests. Just add tests for `"[1:2048, 200:300]"` etc.

mileslucas (changes requested):

src/CCDReduction.jl
siddharthlal25: Right, I don't think it is required for the end user!

src/methods.jl

src/methods.jl
siddharthlal25: For the calculation of `dims` in the case of overscan subtraction.

test/methods.jl
siddharthlal25: Okay, but now I think we either need to change the `fits_indices` function, or the `trimview`. Currently, `trimview` only supports indices of the format `(:, some number)` or `:` at the second place. So, let's modify `trimview` to the case when `(number, number)` (one of the number still remains all rows or all columns, but it will be as range not as `:`) gets processed in it. Thoughts?

src/methods.jl
mileslucas: you shouldn't need this. By the time you call the real `subtract_overscan!` function the indices should already be transformed.

test/methods.jl
mileslucas: that is a good point. I think you should modify trimview to be able to work with `1:100, 80:100` just as well as `:, 80:100`. Let me know if you have issues figuring out how to enable that.

src/methods.jl
siddharthlal25: Interesting, I didn't notice this!

src/methods.jl
```diff
@@ -54,14 +66,17 @@ function subtract_overscan!(frame::AbstractArray, idxs; dims = axes_min_length(i
...
 
 """
     subtract_overscan(frame::AbstractArray, idxs; dims = axes_min_length(idxs))
 
 Subtract the overscan frame from image.
 
 `dims` is the dimension along which `overscan_frame` is combined. The default value
-of `dims` is the axis with smaller length in overscan region.
+of `dims` is the axis with smaller length in overscan region. `idxs` is also supported
+as FITS based indices.
```
```suggestion
of `dims` is the axis with smaller length in overscan region. If `idxs` is a string it will be parsed as FITS-style indices.
```

src/methods.jl
```diff
@@ -134,10 +153,12 @@ flat_correct(frame::AbstractArray, flat_frame::AbstractArray; kwargs...) = flat_
 """
     trim(frame::AbstractArray, idxs)
 
-Trims the frame to remove the region specified by idxs.
+Trims the `frame` to remove the region specified by idxs.
 
 This function trims the array in a manner such that final array should be rectangular.
 The indices follow standard Julia convention, so `(:, 45:60)` trims all columns from 45 to 60 and `(1:20, :)` trims all the rows from 1 to 20.
+The function also supports FITS convention when indices are passed as strings, so `"[45:60, :]"` works equivalently as `(:, 45:60)` and `"[:, 1:20]"`
+works equivalently as `(1:20, :)`.
```
mileslucas: I would keep this briefer- "we support FITS-style indices". That way you can simply address this the same way for the docstrings and in the documentation we can write a section for `Indexing Styles`.

This way you don't have to explain what FITS-style indices are in _every_ docstring that uses them.

mileslucas (approved):

src/methods.jl
```diff
@@ -134,10 +152,11 @@ flat_correct(frame::AbstractArray, flat_frame::AbstractArray; kwargs...) = flat_
 """
     trim(frame::AbstractArray, idxs)
 
-Trims the frame to remove the region specified by idxs.
+Trims the `frame` to remove the region specified by idxs.
 
 This function trims the array in a manner such that final array should be rectangular.
 The indices follow standard Julia convention, so `(:, 45:60)` trims all columns from 45 to 60 and `(1:20, :)` trims all the rows from 1 to 20.
+The function also supports FITS-style indices.
```
giordano, line 159: This docstring doesn't explain what FITS-style indices are, and especially that they must be a string

giordano (commented):
I think I mentioned this already some time ago when discussing the proposal, I'm not exceedingly happy about having the indices as a string to be parsed.  Can we for example have a very simple type wrapping indices instead?

src/methods.jl
siddharthlal25: I have planned to add up a documentation section for Indexing Styles before release. There I will explain about all kinds of indexing styles that we support.

siddharthlal25:
I have no problem in re-writing this for type wrapping indices! I remember you saying strings are evil, what are the troubles that they cause?

giordano:
> what are the troubles that they cause?

That you have to parse them :slightly_smiling_face:  Also, idiomatically a string bears no special meaning, why that should mean "FITS-style indices"?  A simple type would convey that special meaning, and can be better used for dispatch

siddharthlal25:
True, it would be more convenient! Thoughts? @mileslucas

mileslucas:
I don't think it's a huge issue since the fits style indexing is only applicable to `subtract_overscan` and `trim`. However, if we were to do it a different way, I'd consider a viable alternative to be a package FITSIndices.jl that exports a macro like `fits"[1:2048, 2048:2068]"` that gets parsed into Julia indices via a macro so there doesn't have to be an intermediate object.

However, this presents an issue, because the output of the FITS header containing something like "this is the section where the overscan is" is a string, so for us to support using these functions with FITS files directly (#11), we need some way of string parsing internally, still.

giordano:
> However, this presents an issue, because the output of the FITS header containing something like "this is the section where the overscan is" is a string, so for us to support using these functions with FITS files directly, we need some way of string parsing internally, still.

Ok, if the idea is to feed the function directly with a key of a FITS file I guess the string will have its use anyway.

mileslucas:
If we want to avoid having to write/expose functions that take `idxs` as string, we can use the `fits""` macro idea for that, such that internally if we have to parse a string from a header, we do that for the user, and otherwise the parsing happens at compile-time with the macro, avoiding the need for the type specialized functions.

I'm not opposed to that, actually- would be a simple package to set up with just a couple functions and would reduce the number of functions written here.

giordano:
Ok, let's go on with this PR, which anyway already has the parsing in place, and maybe open an issue to remember considering this change in the future

siddharthlal25:
I have opened issue #14 in this regard and I am merging this PR! :tada:

PR #15, opened 2020-05-31
Added dark subtraction

### LINKED ISSUE #12: Dark Subtraction
Add a method `subtract_dark!` (and `subtract_dark`) for dark subtraction. See [this implementation](https://ccdproc.readthedocs.io/en/latest/api/ccdproc.subtract_dark.html#ccdproc.subtract_dark) for an idea of the processing.

This is a little different than the other subtraction routines, because there is a scaling aspect to dark subtraction that can get a little messy depending on units and exposure time. It will be really easy to over-engineer this, so let's try and keep it organized and logical!

I would expect the functions to have a signature like
```julia
subtract_dark(frame, dark_frame; data_exposure=1, dark_exposure=1, scale=false)
```
however, for a version using strings (see #11) there are two fundamental different modes of operation
1. Read files in but don't use keywords
    ```julia
    subtract_dark(filename, filename2; data_exposure=1, dark_exposure=1)
    ```
2. Read files in using keywords
    ```julia
    subtract_dark(filename, filename2; exposure_key)
    ```

Unfortunately it is hard to write different functions based only off their keyword arguments. I have an idea for this but we can discuss it when we get closer to that implementation.

siddharthlal25:
Fixes #12.

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -332,25 +332,22 @@ combine(frames::AbstractVector{<:AbstractArray}; method = median) = combine(fram
 
 
 """
-    subtract_dark!(frame::AbstractArray, dark_frame::AbstractArray; data_exposure = 1, dark_exposure = 1, scale = false)
+    subtract_dark!(frame::AbstractArray, dark_frame::AbstractArray; data_exposure = 1, dark_exposure = 1)
 
-In place version of [`subtract_dark`](@ref)
+In-place version of [`subtract_dark`](@ref)
 
 # See Also
 [`subtract_dark`](@ref)
 """
-function subtract_dark!(frame::AbstractArray, dark_frame::AbstractArray; data_exposure = 1, dark_exposure = 1, scale = false)
-    if scale
-        frame .-= (dark_frame .* (data_exposure / dark_exposure))
-    else
-        frame .-= dark_frame
-    end
+function subtract_dark!(frame::AbstractArray, dark_frame::AbstractArray; data_exposure = 1, dark_exposure = 1)
+    factor = data_exposure / dark_exposure
+    @. frame -= (dark_frame * factor)
     return frame
 end
 
 
 """
-    subtract_dark(frame::AbstractArray, dark_frame::AbstractArray; data_exposure = 1, dark_exposure = 1, scale = false)
+    subtract_dark(frame::AbstractArray, dark_frame::AbstractArray; data_exposure = 1, dark_exposure = 1)
 
 Subtract the `dark_frame` from `frame`.
 
@@ -366,7 +363,7 @@ julia> subtract_dark(frame, dark_frame)
  0.0  0.0  0.0
  0.0  0.0  0.0
 
-julia> subtract_dark(frame, dark_frame, data_exposure = 1, dark_exposure = 4, scale = true)
+julia> subtract_dark(frame, dark_frame, data_exposure = 1, dark_exposure = 4)
 3×3 Array{Float64,2}:
  0.75  0.75  0.75
  0.75  0.75  0.75
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -329,3 +329,52 @@ function combine(frames::Vararg{<:AbstractArray}; method = median)
 end
 
 combine(frames::AbstractVector{<:AbstractArray}; method = median) = combine(frames..., method = method)
+
+
+"""
+    subtract_dark!(frame::AbstractArray, dark_frame::AbstractArray; data_exposure = 1, dark_exposure = 1, scale = false)
+
+In place version of [`subtract_dark`](@ref)
+
+# See Also
+[`subtract_dark`](@ref)
+"""
+function subtract_dark!(frame::AbstractArray, dark_frame::AbstractArray; data_exposure = 1, dark_exposure = 1, scale = false)
+    if scale
+        frame .-= (dark_frame .* (data_exposure / dark_exposure))
+    else
+        frame .-= dark_frame
+    end
+    return frame
+end
+
+
+"""
+    subtract_dark(frame::AbstractArray, dark_frame::AbstractArray; data_exposure = 1, dark_exposure = 1, scale = false)
+
+Subtract the `dark_frame` from `frame`.
+
+# Examples
+```jldoctest
+julia> frame = ones(3, 3);
+
+julia> dark_frame = ones(3, 3);
+
+julia> subtract_dark(frame, dark_frame)
+3×3 Array{Float64,2}:
+ 0.0  0.0  0.0
+ 0.0  0.0  0.0
+ 0.0  0.0  0.0
+
+julia> subtract_dark(frame, dark_frame, data_exposure = 1, dark_exposure = 4, scale = true)
+3×3 Array{Float64,2}:
+ 0.75  0.75  0.75
+ 0.75  0.75  0.75
+ 0.75  0.75  0.75
+
+```
+
+# See Also
+[`subtract_dark!`](@ref)
+"""
+subtract_dark(frame::AbstractArray, dark_frame::AbstractArray; kwargs...) = subtract_dark!(deepcopy(frame), dark_frame; kwargs...)
--- a/test/methods.jl
+++ b/test/methods.jl
@@ -130,7 +130,7 @@ end
     # testing non-mutating version
     @test @inferred(subtract_dark(ones(5, 5), ones(5, 5))) == zeros(5, 5)
     @test @inferred(subtract_dark(ones(5, 5), ones(5, 5), data_exposure = 1, dark_exposure = 4)) == fill(0.75, 5, 5)
-    @test @inferred(subtract_dark(ones(5, 5), ones(5, 5), data_exposure = 13, dark_exposure = 17)) ≈ fill(Float32(4/17), 5, 5)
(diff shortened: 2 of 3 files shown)


test/methods.jl
```diff
@@ -126,6 +126,26 @@ end
...
+@testset "dark subtraction" begin
+    # testing non-mutating version
+    @test @inferred(subtract_dark(ones(5, 5), ones(5, 5))) == zeros(5, 5)
+    @test @inferred(subtract_dark(ones(5, 5), ones(5, 5), data_exposure = 1, dark_exposure = 4, scale = true)) == fill(0.75, 5, 5)
+
+    # testing mutating version
+    frame = ones(5, 5)
+    dark_frame = ones(5, 5)
+    subtract_dark!(frame, dark_frame)
+    @test frame ≈ zeros(5, 5)
```
siddharthlal25: @giordano @miles if the above line is changed to `@test frame == zeros(5, 5)`, it fails. Any idea why is it so?

siddharthlal25:
Should we keep a check for `dark_exposure > 0`? @mileslucas @giordano

test/methods.jl
mileslucas: It might be from floating point round off

src/methods.jl
```diff
@@ -329,3 +329,52 @@ function combine(frames::Vararg{<:AbstractArray}; method = median)
...
+    subtract_dark!(frame::AbstractArray, dark_frame::AbstractArray; data_exposure = 1, dark_exposure = 1, scale = false)
+
+In place version of [`subtract_dark`](@ref)
+
+# See Also
+[`subtract_dark`](@ref)
+"""
+function subtract_dark!(frame::AbstractArray, dark_frame::AbstractArray; data_exposure = 1, dark_exposure = 1, scale = false)
+    if scale
+        frame .-= (dark_frame .* (data_exposure / dark_exposure))
```
giordano: Why isn't this broadcasting everything?  You're creating a temporary array.  Doesn't
```suggestion
        @. frame -= (dark_frame * (data_exposure / dark_exposure))
```
work as expected?

**Edit**: ok, it's probably not creating a temporary array, but broadcasting everything tends to be a good idea anyway.

Also, is the `scale` argument necessary?  I think you can just always use this line.  If one wants to scale, they have to remember to set `scale = true` in addition to setting the exposure, it's a bad gotcha

mileslucas: On second look yes having scale be an argument seems to be a remnant of ancient iraf code. I agree it would be better to skip it.

giordano:
@siddharthlal25 I've noted that you always use the same branch called `devel` for your work.  I just wanted to let you know that it isn't a good practice to do so, for each feature you should create _a new branch_ (possibly with a somewhat short yet meaningful name, easy for you to identify).  Using a single branch workflow doesn't let you work on multiple features at the same time, and I think it's a good idea to practice with git's features :wink:   (Also, the fact that you always use the same branch doesn't make my life particularly easy when I want to check out your work locally)

mileslucas:
> Should we keep a check for `dark_exposure > 0`? @mileslucas @giordano 

No, I think this is clear enough that we don’t need to error check it

test/methods.jl
giordano: Why does it fail for you?
```julia
julia> frame = ones(5, 5)
5×5 Array{Float64,2}:
 1.0  1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0  1.0

julia> dark_frame = ones(5, 5)
5×5 Array{Float64,2}:
 1.0  1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0  1.0

julia> subtract_dark!(frame, dark_frame)
5×5 Array{Float64,2}:
 0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0

julia> frame == zeros(5, 5)
true
```

src/methods.jl
```diff
@@ -329,3 +329,52 @@ function combine(frames::Vararg{<:AbstractArray}; method = median)
 end
 
 combine(frames::AbstractVector{<:AbstractArray}; method = median) = combine(frames..., method = method)
+
+
+"""
+    subtract_dark!(frame::AbstractArray, dark_frame::AbstractArray; data_exposure = 1, dark_exposure = 1, scale = false)
+
+In place version of [`subtract_dark`](@ref)
```
giordano:
```suggestion
In-place version of [`subtract_dark`](@ref)
```

siddharthlal25: > @siddharthlal25 I've noted that you always use the same branch called `devel` for your work. I just wanted to let you know that it isn't a good practice to do so, for each feature you should create _a new branch_ (possibly with a somewhat short yet meaningful name, easy for you to identify). Using a single branch workflow doesn't let you work on multiple features at the same time, and I think it's a good idea to practice with git's features  (Also, the fact that you always use the same branch doesn't make my life particularly easy when I want to check out your work locally)

Sure, I will keep something short and meaningful! Sorry for the hassle.

siddharthlal25:
> @siddharthlal25 I've noted that you always use the same branch called `devel` for your work. I just wanted to let you know that it isn't a good practice to do so, for each feature you should create _a new branch_ (possibly with a somewhat short yet meaningful name, easy for you to identify). Using a single branch workflow doesn't let you work on multiple features at the same time, and I think it's a good idea to practice with git's features  (Also, the fact that you always use the same branch doesn't make my life particularly easy when I want to check out your work locally)

Sure, I will keep something short and meaningful! Sorry for the hassle.

src/methods.jl
siddharthlal25: Sure, I will make the changes!

test/methods.jl
siddharthlal25: I don't know what went wrong at that time, it is working fine right now!

giordano: Yeah, I'd be surprised if you don't get zero when simply subtracting two identical matrices, without further operations

src/methods.jl
```diff
@@ -329,3 +329,49 @@ function combine(frames::Vararg{<:AbstractArray}; method = median)
...
+    subtract_dark!(frame::AbstractArray, dark_frame::AbstractArray; data_exposure = 1, dark_exposure = 1)
+
+In-place version of [`subtract_dark`](@ref)
+
+# See Also
+[`subtract_dark`](@ref)
+"""
+function subtract_dark!(frame::AbstractArray, dark_frame::AbstractArray; data_exposure = 1, dark_exposure = 1)
+    factor = data_exposure / dark_exposure
+    @. frame -= (dark_frame * factor)
```
siddharthlal25, line 344: I don't know whether broadcasting is optimising repetitive calculation or not, so I switched to calculate that multiplying factor one step before and then use it next. I think this will save us some extra divisions. Thoughts?

giordano: This might be fine, even though I'm a bit worried about type stability.  What happens if `frame` is a `Float32`?

giordano:
I think in general we should have more tests about type stability with types different from `Float64`.  These are very much common in this field, so it's important to make sure that everything works smoothly in these cases.

src/methods.jl
giordano: >  What happens if frame is a `Float32`?

Ok, it seems to work fine:
```julia
julia> A = rand(Float32, 3, 3);

julia> B = rand(Float32, 3, 3);

julia> subtract_dark!(A, B)
3×3 Array{Float32,2}:
  0.352384  -0.14378    0.698295
 -0.394064   0.184986  -0.0990469
 -0.676293   0.786632   0.0116163
```

siddharthlal25: Tried this,
```
frame = rand(Float32, 5, 5)
dark_frame = rand(Float32, 5, 5)
@inferred(subtract_dark!(frame, dark_frame, data_exposure = 7, dark_exposure = 13))
```
This did not error!

siddharthlal25: These comments are quite unsynced!:laughing:

test/methods.jl
```diff
@@ -126,6 +126,32 @@ end
     @test_throws ArgumentError combine()
 end
 
+@testset "dark subtraction" begin
+    # testing non-mutating version
+    @test @inferred(subtract_dark(ones(5, 5), ones(5, 5))) == zeros(5, 5)
+    @test @inferred(subtract_dark(ones(5, 5), ones(5, 5), data_exposure = 1, dark_exposure = 4)) == fill(0.75, 5, 5)
+    @test @inferred(subtract_dark(ones(5, 5), ones(5, 5), data_exposure = 13, dark_exposure = 17)) ≈ fill(Float32(4/17), 5, 5)
```
giordano: I'm not sure how this one is useful for more general testing type-stability, since the arguments of `subtract_dark` are `Array{Float64}`

siddharthlal25: I missed `Float32` over there. :sweat_smile:

mileslucas (approved):
Looks good!

giordano (approved):

PR #16, opened 2020-06-03
Added interface for FITS files for subtract overscan

### LINKED ISSUE #11: Add interface for FITS files

## Example usage
```julia
# or even
```

### FITS files
```julia
# load HDUS
data = read(hdu)
hdr = read_header(hdu)
```

```julia
# use header key
```
### Filename strings

```julia
```

### Code Generation
```julia
end
```

```
"""
    cropview(frame, idxs)
"""
```

### Special Cases
1. function takes two frames

2. Getting indices from file
	```
	```
	```julia
    # equivalent to
    hdus = FITS("filename.fits")
    frame = read(hdus[1])
    hdr = read_header(hdus[1])
    idxs = hdr[string(:BIASSEC)]
    ```

siddharthlal25:
This PR is a part to fix #11. It adds interface for `fits` files for overscan subtraction.


src/methods.jl
```diff
@@ -13,6 +13,9 @@ function fits_indices(string::String)
     return reverse(idxs)
 end
 
+# Converts Int to Float and leaves Float to Float itself
+convert_value(x, S) = S <: Integer ? round(S, x) : convert(S, x)
```
giordano: Why not defining two methods
```suggestion
convert_value(x, S::Integer) = round(S, x)
convert_value(x, S) = convert(S, x)
```
?

Note also that traditionally, the type argument is the first one, as you can see in the `round` and `convert` methods you're calling here

src/methods.jl
```diff
@@ -59,15 +62,20 @@ In-place version of [`subtract_overscan`](@ref)
...
     frame .-= overscan_value
     return frame
 end
 
 subtract_overscan!(frame::AbstractArray, idxs::String; kwargs...) = subtract_overscan!(frame, fits_indices(idxs); kwargs...)
 
+# support for frame as strings(Location of file) and ImageHDU
+subtract_overscan!(frame::ImageHDU, key::Symbol; kwargs...) = subtract_overscan!(read(frame), read_header(frame)[string(key)]; kwargs...)
+subtract_overscan!(frame::ImageHDU, idxs; kwargs...) = subtract_overscan!(read(frame), idxs; kwargs...)
+subtract_overscan!(filename::String, idxs; hdu = 1, kwargs...) = subtract_overscan!(FITS(filename)[hdu], idxs; kwargs...)
```
giordano: Missing something here? :wink:

src/methods.jl
mileslucas: Yes, much better than my suggestion! I guess late-night MIles likes writing Julia anti-patterns 😅

src/methods.jl
giordano: Uh, looking better, probably not!

mileslucas:
Argh, something I was hoping we could avoid dealing with turns out we have to deal with it. 

FITSIO.jl just takes over the memory read in by cfitsio. cfitsio reads the data in as row-major, but when Julia takes that memory it is assumed to be column-major. This means all of the data read in by `read(::ImageHDU)` is transposed. This will cause issues with any method that requires indexing (overscan, trimming, cropping). For example, I just tried subtracting overscan from a file using its fits header and I got an IndexError.

since we are handling opening the files internally, we're going to have to permute the dimensions appropriately. I wanted to avoid this because it is discrepant from FITSIO.jl (although the FITSIO.jl behavior isn't what I'd like it to be). 

I suggest we add a function like this:
```julia
function getdata(hdu::ImageHDU)
	data = read(hdu)
	d = ndims(data)
	return permutedims(data, d:-1:1)
end
```

and the further calls within our methods should look like
```julia
subtract_overscan(hdu::ImageHDU, idxs; kwargs...) =
	subtract_overscan!(getdata(hdu), idxs; kwargs...)
```

src/methods.jl
siddharthlal25: > Missing something here?

Are you referring to dispatch for `frame` as `FITS`?

giordano: No, the semicolon: I didn't see it before `kwargs`, I was it was missing, then I realised there was another keyword argument before it :sweat_smile:

siddharthlal25: I think we should add the dispatch for `FITS` as well, if we say that `FITS` is very broad and it's not good to narrow it down just for our case, we are already doing that when we are writing a `String` dispatch for `frame`. Thoughts?

siddharthlal25:
> Argh, something I was hoping we could avoid dealing with turns out we have to deal with it.
> 
> FITSIO.jl just takes over the memory read in by cfitsio. cfitsio reads the data in as row-major, but when Julia takes that memory it is assumed to be column-major. This means all of the data read in by `read(::ImageHDU)` is transposed. This will cause issues with any method that requires indexing (overscan, trimming, cropping). For example, I just tried subtracting overscan from a file using its fits header and I got an IndexError.
> 
> since we are handling opening the files internally, we're going to have to permute the dimensions appropriately. I wanted to avoid this because it is discrepant from FITSIO.jl (although the FITSIO.jl behavior isn't what I'd like it to be).
> 
> I suggest we add a function like this:
> 
> ```julia
> function getdata(hdu::ImageHDU)
> 	data = read(hdu)
> 	d = ndims(data)
> 	return permutedims(data, d:-1:1)
> end
> ```
> 
> and the further calls within our methods should look like
> 
> ```julia
> subtract_overscan(hdu::ImageHDU, idxs; kwargs...) =
> 	subtract_overscan!(getdata(hdu), idxs; kwargs...)
> ```

Can't we just return the transposed version of data, something like:
```
function getdata(hdu::ImageHDU)
    data = read(hdu)
    return data'
end
```
Edit:
The return type here becomes `LinearAlgebra.Adjoint{Int64,Array{Int64,2}}`, whereas for the version you suggested the return type is `Array{Int64,2}`. Let's keep our types as simple as possible, so let's go with your version.

src/methods.jl
siddharthlal25: Switched to conventional argument style, and the functions look something like:
```
convert_value(S::Integer, x) = round(S, x)
convert_value(S, x) = convert(S, x)
```
The function call `convert_value(Int16, 5.4)` goes to the second function, whereas it should have gone to the first one, why is it happening? :thinking:

mileslucas: haha, we both had a folly-
```julia
convert_value(S::Type{<:Integer}, x)
```

src/methods.jl
mileslucas: People don't really work with FITS objects directly, but with the HDUs instead. Plus this would just add another piece of complexity.

mileslucas:
> LinearAlgebra.Adjoint{Int64,Array{Int64,2}}, whereas for the version you suggested the return type is Array{Int64,2}

Precisely, plus you can't use literal `'` (or `transpose`) for anything more than 2 dimensions.

.gitignore
```diff
@@ -22,6 +22,9 @@ deps/usr/
 deps/src/
 deps/*.fits
 
+# No FITS file present in tests directory to be included
+test/*.fits
```
mileslucas: we can very safely just put `*.fits` to ignore any fits file in any folder

src/methods.jl
```diff
@@ -92,18 +105,20 @@ julia> subtract_overscan(frame, "[4:5, 1:1]", dims = 2)
...
 
 # See Also
-* [`subtract_overscan!`](@ref)
+[`subtract_overscan!`](@ref)
 """
-subtract_overscan(frame::AbstractArray, idxs; dims = axes_min_length(idxs)) = subtract_overscan!(deepcopy(frame), idxs, dims = dims)
-
+subtract_overscan(frame, idxs; kwargs...) = subtract_overscan!(deepcopy(frame), idxs; kwargs...)
+subtract_overscan(frame::ImageHDU, idxs; kwargs...) = subtract_overscan!(getdata(frame), idxs; kwargs...)
+subtract_overscan(frame::ImageHDU, key::Symbol; kwargs...) = subtract_overscan!(getdata(frame), read_header(frame)[string(key)]; kwargs...)
```
mileslucas: it doesn't matter right now because the behavior is identical, but you could have also used this version, which dispatches to the function a line above this.  
```suggestion
subtract_overscan(frame::ImageHDU, key::Symbol; kwargs...) = subtract_overscan(frame, read_header(frame)[string(key)]; kwargs...)
```

test/runtests.jl
```diff
@@ -1,5 +1,11 @@
 using CCDReduction
 using Test
 using Statistics
+using FITSIO
+
+const test_file_url_M6707HH = "https://www.cfa.harvard.edu/~dfabricant/huchra/ay16/fits/M6707HH.fits"
+const test_file_path_M6707HH = joinpath(@__DIR__, "MH6707HH.fits")
+!isfile(test_file_path_M6707HH) && download(test_file_url_M6707HH, test_file_path_M6707HH)
```
mileslucas: For the CI to cache we should give it a _directory_. we don't want to have it cache our entire test directory, though, so I suggest this
```suggestion
const test_file_path_M6707HH = joinpath(@__DIR__, "data", "MH6707HH.fits")
if !isfile(test_file_path_M6707HH)
	mkpath(dirname(test_file_path_M6707HH))
	download(test_file_url_M6707HH, test_file_path_M6707HH)
end
```

mileslucas: Also, you need to add the cache directive to travis:
```
cache:
    directories:
		- tests/data
```

mileslucas (changes requested):

src/methods.jl
```diff
@@ -13,6 +13,17 @@ function fits_indices(string::String)
     return reverse(idxs)
 end
 
+# Converts Int to Float and leaves Float to Float itself
```
```suggestion
# Convert using `round` for integers
```

src/methods.jl
```diff
@@ -92,9 +103,12 @@ julia> subtract_overscan(frame, "[4:5, 1:1]", dims = 2)
 ```
 
 # See Also
-* [`subtract_overscan!`](@ref)
+[`subtract_overscan!`](@ref)
 """
-subtract_overscan(frame::AbstractArray, idxs; dims = axes_min_length(idxs)) = subtract_overscan!(deepcopy(frame), idxs, dims = dims)
+subtract_overscan(frame, idxs; kwargs...) = subtract_overscan!(deepcopy(frame), idxs; kwargs...)
+subtract_overscan(frame::ImageHDU, idxs; kwargs...) = subtract_overscan!(getdata(frame), idxs; kwargs...)
```
mileslucas, line 116: Please add a docstring here
```julia
"""
	subtract_overscan(::FITSIO.ImageHDU, idxs; [dims])
	subtract_overscan(filename, idxs; hdu=1, [dims])

Load a FITS file or HDU before subtracting the overscan region. If `idxs` is a symbol it will be read from the FITS header with that key (case sensitive).
"""
```

mileslucas (changes requested):

mileslucas (approved):

giordano (approved):

PR #17, opened 2020-06-05
Added fits interface for trim and crop

### LINKED ISSUE #11: Add interface for FITS files

siddharthlal25:
Partially fixes #11, will switch to codegen when all code written separately works in a correct manner.

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -212,9 +212,10 @@ trim(frame::AbstractArray, idxs) = copy(trimview(frame, idxs))
     trim(::FITSIO.ImageHDU, idxs)
     trim(filename, idxs; hdu=1)
 
-Load a FITS file or HDU before trimming.
+Load a FITS file or HDU before trimming. If `idxs` is a symbol it will be read from the FITS header with that key (case sensitive).
 """
 trim(frame::ImageHDU, idxs) = trimview(getdata(frame), idxs)
+trim(frame::ImageHDU, idxs::Symbol) = trim(frame, read_header(frame)[string(idxs)])
 trim(frame::String, idxs; hdu = 1) = trim(FITS(frame)[hdu], idxs)
 
 
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -295,7 +295,7 @@ crop(frame, shape; kwargs...) = copy(cropview(frame, shape; kwargs...))
     crop(::FITSIO.ImageHDU, shape; force_equal = true)
     crop(filename, shape; hdu=1, force_equal = true)
 
-Load a FITS file or HDU before trimming.
+Load a FITS file or HDU before cropping.
 """
 crop(frame::ImageHDU, shape; kwargs...) = crop(getdata(frame), shape; kwargs...)
 crop(filename::String, shape; hdu = 1, kwargs...) = crop(FITS(filename)[hdu], shape; kwargs...)
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -107,6 +107,7 @@ julia> subtract_overscan(frame, "[4:5, 1:1]", dims = 2)
 """
 subtract_overscan(frame, idxs; kwargs...) = subtract_overscan!(deepcopy(frame), idxs; kwargs...)
 
+
 """
 	subtract_overscan(::FITSIO.ImageHDU, idxs; [dims])
 	subtract_overscan(filename, idxs; hdu=1, [dims])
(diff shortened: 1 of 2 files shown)


src/methods.jl
```diff
@@ -207,6 +208,16 @@ julia> trim(frame, "[2:5, 1:5]")
...
 
 
+"""
+    trim(::FITSIO.ImageHDU, idxs)
+    trim(filename, idxs; hdu=1)
+
+Load a FITS file or HDU before trimming.
+"""
+trim(frame::ImageHDU, idxs) = trimview(getdata(frame), idxs)
+trim(frame::String, idxs; hdu = 1) = trim(FITS(frame)[hdu], idxs)
```
mileslucas, line 218: please add a method (and relevant documentation) for getting the idxs as a header key. This should be exactly the same as `subtract_overscan`.

mileslucas (changes requested):
For this method, since `trimview` doesn't alter the data, you should do the following:

change
```julia
trim(frame::AbstractArray, idxs) = copy(trimview(frame, idxs))
```
to
```julia
trim(frame, idxs; kwargs...) = copy(trimview(frame, idxs; kwargs...))
```

and then add the methods
```julia
trimview(::AbstractArray, idxs)
trimview(::ImageHDU, idxs)
trimview(::ImageHDU, key::Symbol)
trimview(::String, idxs; hdu=1)
```

siddharthlal25:
`trimview` is a view of the array, and in documentation for `trimview` we have mentioned that subsequent changes to `frame` will lead to change in trimmed version. Whereas in this case when we provide `frame` as `ImageHDU`, mutation (subsequent mutation in `ImageHDU` won't alter the trimmed version) won't take place, that's why I think `trim` would be the right place to define the methods. Thoughts?

Edit:
Also, we shouldn't add any methods to `trimview`, the same as we did for `subtract_overscan!`, because we can't mutate the `ImageHDU` or files as of now.
PS: These are my thoughts, I would like to know your viewpoint.

Edit Again:
Suppose we define `trimview`, we will be using `getdata` to get the data (finally as array), and when we read files (using `read`, it allocates memory because it copies (I think this)), so suppose someone calls `trim` with `ImageHDU` so there will be two copies, one because of this:
```
trim(frame, idxs; kwargs...) = copy(trimview(frame, idxs; kwargs...))
```
and second in the `getdata` in next phase.

Edit Again Again :rofl: 
I should have added the method with `key::Symbol`, I will add them after we resolve this!

mileslucas:
> Also, we shouldn't add any methods to trimview, the same as we did for subtract_overscan!, because we can't mutate the ImageHDU or files as of now.

`trimview` is not a mutating function

> Also, we shouldn't add any methods to trimview, the same as we did for subtract_overscan!, because we can't mutate the ImageHDU or files as of now.

That is a fair point. Thinking a little bit harder my original reasoning behind allowing it is not as strong. So yes, only provide trim functions (this should look _extremely_ similar to subtract_overscan)

siddharthlal25:
I would prefer doing `subtract_bias` and `correct_flat` in separate PR's so that things don't mess up. :smiley: 
Edit: 
Or maybe we can do `subtract_bias` and `correct_flat` in one, and `subtract_dark` in another one. Thoughts?

src/methods.jl
```diff
@@ -276,7 +288,17 @@ julia> crop(frame, (4, 3), force_equal = false)
...
 """
-crop(frame::AbstractArray, shape; kwargs...) = copy(cropview(frame, shape; kwargs...))
+crop(frame, shape; kwargs...) = copy(cropview(frame, shape; kwargs...))
+
+
+"""
+    crop(::FITSIO.ImageHDU, shape; force_equal = true)
+    crop(filename, shape; hdu=1, force_equal = true)
+
+Load a FITS file or HDU before trimming.
```
```suggestion
Load a FITS file or HDU before cropping.
```

mileslucas (changes requested):

mileslucas (approved):

giordano (approved):

PR #18, opened 2020-06-06
Added FITS interface for subtract_bias and flat_correct

### LINKED ISSUE #11: Add interface for FITS files

siddharthlal25:
Added FITS interface for `subtract_bias` and `flat_correct`.

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -61,6 +61,21 @@ julia> subtract_bias(frame, bias)
 """
 subtract_bias(frame::AbstractArray, bias_frame::AbstractArray) = subtract_bias!(deepcopy(frame), bias_frame)
 
+"""
+    subtract_bias(::Union{FITSIO.ImageHDU, AbstractArray}, ::Union{FITSIO.ImageHDU, AbstractArray})
+	subtract_bias(::Union{FITSIO.ImageHDU, AbstractArray}, ::String; hdu_bias = 1)
+	subtract_bias(::String, ::Union{FITSIO.ImageHDU, AbstractArray}; hdu_frame = 1)
+	subtract_bias(::String, ::String; hdu_frame = 1, hdu_bias = 1)
+
+Load a FITS file or HDU before bias subtraction.
+"""
+subtract_bias(frame::ImageHDU, bias_frame::AbstractArray) = subtract_bias(getdata(frame), bias_frame)
+subtract_bias(frame::AbstractArray, bias_frame::ImageHDU) = subtract_bias(frame, getdata(bias_frame))
+subtract_bias(frame::ImageHDU, bias_frame::ImageHDU) = subtract_bias(getdata(frame), bias_frame)
+subtract_bias(frame::String, bias_frame; hdu_frame = 1) = subtract_bias(FITS(frame)[hdu_frame], bias_frame)
+subtract_bias(frame, bias_frame::String; hdu_bias = 1) = subtract_bias(frame, FITS(bias_frame)[hdu_bias])
+subtract_bias(frame::String, bias_frame::String; hdu_frame = 1, hdu_bias = 1) = subtract_bias(frame, FITS(bias_frame)[hdu_bias]; hdu_frame = hdu_frame)
+
 
 """
     subtract_overscan!(frame::AbstractArray, idxs; dims = axes_min_length(idxs))
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -41,12 +41,12 @@ end
 
 """
     subtract_bias!(frame::AbstractArray, bias_frame::ImageHDU)
-    subtract_bias!(frame::AbstractArray, bias_frame::String; hdu_bias = 1)
+    subtract_bias!(frame::AbstractArray, bias_frame::String; hdu = 1)
 
-Load a FITS file or HDU before mutating bias subtraction. The `frame` must be of the type `AbstractArray`.
+Load a FITS file or HDU for the bias frame before subtracting from `frame` in-place.
 """
 subtract_bias!(frame::AbstractArray, bias_frame::ImageHDU) = subtract_bias!(frame, getdata(bias_frame))
-subtract_bias!(frame::AbstractArray, bias_frame::String; hdu_bias = 1) = subtract_bias!(frame, FITS(bias_frame)[hdu_bias])
+subtract_bias!(frame::AbstractArray, bias_frame::String; hdu = 1) = subtract_bias!(frame, FITS(bias_frame)[hdu])
 
 
 """
@@ -73,19 +73,21 @@ subtract_bias(frame::AbstractArray, bias_frame::AbstractArray) = subtract_bias!(
 
 
 """
-    subtract_bias(::Union{FITSIO.ImageHDU, AbstractArray}, ::Union{FITSIO.ImageHDU, AbstractArray})
-    subtract_bias(::Union{FITSIO.ImageHDU, AbstractArray}, ::String; hdu_bias = 1)
-    subtract_bias(::String, ::Union{FITSIO.ImageHDU, AbstractArray}; hdu_frame = 1)
-    subtract_bias(::String, ::String; hdu_frame = 1, hdu_bias = 1)
+    subtract_bias(frame, bias_frame; [hdu = 1])
 
-Load a FITS file or HDU before bias subtraction.
+Subtract the bias frame from `frame`. If either arguments are `FITSIO.ImageHDU` they will be loaded into memory.
+If either arguments are strings we will attempt to locate a FITS file and open it before loading the data from the given `hdu`.
+If loading multiple files, you can specify the HDU numbers separately (`hdu=(1, 2)`) or simultanesously (`hdu=1`).
 """
 subtract_bias(frame::ImageHDU, bias_frame::AbstractArray) = subtract_bias(getdata(frame), bias_frame)
 subtract_bias(frame::AbstractArray, bias_frame::ImageHDU) = subtract_bias(frame, getdata(bias_frame))
 subtract_bias(frame::ImageHDU, bias_frame::ImageHDU) = subtract_bias(getdata(frame), bias_frame)
-subtract_bias(frame::String, bias_frame; hdu_frame = 1) = subtract_bias(FITS(frame)[hdu_frame], bias_frame)
-subtract_bias(frame, bias_frame::String; hdu_bias = 1) = subtract_bias(frame, FITS(bias_frame)[hdu_bias])
-subtract_bias(frame::String, bias_frame::String; hdu_frame = 1, hdu_bias = 1) = subtract_bias(frame, FITS(bias_frame)[hdu_bias]; hdu_frame = hdu_frame)
+subtract_bias(frame::String, bias_frame; hdu = 1) = subtract_bias(FITS(frame)[hdu], bias_frame)
+subtract_bias(frame, bias_frame::String; hdu = 1) = subtract_bias(frame, FITS(bias_frame)[hdu])
+function subtract_bias(frame::String, bias_frame::String; hdu = (1, 1))
+	hdus = hdu isa Integer ? (hdu, hdu) : hdu
+	return subtract_bias(FITS(frame)[hdu[1]], FITS(bias_frame)[hdus[2]])
+end
 
 
 """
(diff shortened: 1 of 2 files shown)


src/methods.jl
```diff
@@ -39,6 +39,16 @@ function subtract_bias!(frame::AbstractArray, bias_frame::AbstractArray)
 end
 
 
+"""
+    subtract_bias!(frame::AbstractArray, bias_frame::ImageHDU)
```
siddharthlal25, line 43: I don't know what's going wrong here, it looks perfectly fine in my editor. (same on line 76) Is it a GH glitch?

mileslucas: I don't see what you're talking about.

siddharthlal25: Fixed it! It was an indentation error from my end.

siddharthlal25:
Let me know your thoughts about these constructors @giordano @mileslucas. Once they are good to go, I will add the constructors for `correct_flat`!

src/methods.jl
```diff
@@ -39,6 +39,16 @@ function subtract_bias!(frame::AbstractArray, bias_frame::AbstractArray)
...
+    subtract_bias!(frame::AbstractArray, bias_frame::String; hdu_bias = 1)
+
+Load a FITS file or HDU before mutating bias subtraction. The `frame` must be of the type `AbstractArray`.
```
```suggestion
Load a FITS file or HDU for the bias frame before subtracting from `frame` in-place.
```

src/methods.jl
```diff
@@ -62,6 +72,22 @@ julia> subtract_bias(frame, bias)
 subtract_bias(frame::AbstractArray, bias_frame::AbstractArray) = subtract_bias!(deepcopy(frame), bias_frame)
 
 
+"""
+    subtract_bias(::Union{FITSIO.ImageHDU, AbstractArray}, ::Union{FITSIO.ImageHDU, AbstractArray})
+    subtract_bias(::Union{FITSIO.ImageHDU, AbstractArray}, ::String; hdu_bias = 1)
+    subtract_bias(::String, ::Union{FITSIO.ImageHDU, AbstractArray}; hdu_frame = 1)
+    subtract_bias(::String, ::String; hdu_frame = 1, hdu_bias = 1)
```
mileslucas: At this point we are making the documentation confusing by writing out too many methods and using awkward unions.

```suggestion
    subtract_bias(frame, bias_frame; [hdu=1])

Subtract the bias frame from `frame`. If either arguments are `FITSIO.ImageHDU` they will be loaded into memory. If either arguments are strings we will attempt to locate a FITS file and open it before loading the data from the given `hdu`. If loading multiple files, you can specify the HDU numbers separately (`hdu=(1, 2)`) or simultanesously (`hdu=1`).
```

src/methods.jl
```diff
@@ -62,6 +72,22 @@ julia> subtract_bias(frame, bias)
...
+
+Load a FITS file or HDU before bias subtraction.
+"""
+subtract_bias(frame::ImageHDU, bias_frame::AbstractArray) = subtract_bias(getdata(frame), bias_frame)
+subtract_bias(frame::AbstractArray, bias_frame::ImageHDU) = subtract_bias(frame, getdata(bias_frame))
+subtract_bias(frame::ImageHDU, bias_frame::ImageHDU) = subtract_bias(getdata(frame), bias_frame)
+subtract_bias(frame::String, bias_frame; hdu_frame = 1) = subtract_bias(FITS(frame)[hdu_frame], bias_frame)
+subtract_bias(frame, bias_frame::String; hdu_bias = 1) = subtract_bias(frame, FITS(bias_frame)[hdu_bias])
+subtract_bias(frame::String, bias_frame::String; hdu_frame = 1, hdu_bias = 1) = subtract_bias(frame, FITS(bias_frame)[hdu_bias]; hdu_frame = hdu_frame)
```
mileslucas: I think a more consistent interface might look like
```suggestion
subtract_bias(frame::ImageHDU, bias_frame::AbstractArray) = subtract_bias(getdata(frame), bias_frame)
subtract_bias(frame::AbstractArray, bias_frame::ImageHDU) = subtract_bias(frame, getdata(bias_frame))
subtract_bias(frame::ImageHDU, bias_frame::ImageHDU) = subtract_bias(getdata(frame), bias_frame)
subtract_bias(frame::String, bias_frame; hdu = 1) = subtract_bias(FITS(frame)[hdu], bias_frame)
subtract_bias(frame, bias_frame::String; hdu = 1) = subtract_bias(frame, FITS(bias_frame)[hdu])
function subtract_bias(frame::String, bias_frame::String; hdu = (1, 1))
	hdus = hdu isa Integer ? (hdu, hdu) : hdu
	subtract_bias(FITS(frame)[hdus[1]], FITS(bias_frame)[hdus[2]])
end
```

src/methods.jl
```diff
@@ -39,6 +39,16 @@ function subtract_bias!(frame::AbstractArray, bias_frame::AbstractArray)
...
 
 
+"""
+    subtract_bias!(frame::AbstractArray, bias_frame::ImageHDU)
+    subtract_bias!(frame::AbstractArray, bias_frame::String; hdu_bias = 1)
+
+Load a FITS file or HDU before mutating bias subtraction. The `frame` must be of the type `AbstractArray`.
+"""
+subtract_bias!(frame::AbstractArray, bias_frame::ImageHDU) = subtract_bias!(frame, getdata(bias_frame))
+subtract_bias!(frame::AbstractArray, bias_frame::String; hdu_bias = 1) = subtract_bias!(frame, FITS(bias_frame)[hdu_bias])
```
```suggestion
subtract_bias!(frame::AbstractArray, bias_frame::String; hdu = 1) = subtract_bias!(frame, FITS(bias_frame)[hdu])
```

mileslucas (changes requested):

mileslucas:
> Partially fixes #11.

Do you mind removing this? There's no point in automatically closing the issue just to reopen it again.

src/methods.jl
```diff
@@ -62,6 +72,24 @@ julia> subtract_bias(frame, bias)
...
+If loading multiple files, you can specify the HDU numbers separately (`hdu=(1, 2)`) or simultanesously (`hdu=1`).
+"""
+subtract_bias(frame::ImageHDU, bias_frame::AbstractArray) = subtract_bias(getdata(frame), bias_frame)
+subtract_bias(frame::AbstractArray, bias_frame::ImageHDU) = subtract_bias(frame, getdata(bias_frame))
+subtract_bias(frame::ImageHDU, bias_frame::ImageHDU) = subtract_bias(getdata(frame), bias_frame)
+subtract_bias(frame::String, bias_frame; hdu = 1) = subtract_bias(FITS(frame)[hdu], bias_frame)
+subtract_bias(frame, bias_frame::String; hdu = 1) = subtract_bias(frame, FITS(bias_frame)[hdu])
+function subtract_bias(frame::String, bias_frame::String; hdu = (1, 1))
+	hdus = hdu isa Integer ? (hdu, hdu) : hdu
+	return subtract_bias(FITS(frame)[hdu[1]], FITS(bias_frame)[hdus[2]])
```
```suggestion
	hdus = hdu isa Integer ? (hdu, hdu) : hdu
	return subtract_bias(FITS(frame)[hdus[1]], FITS(bias_frame)[hdus[2]])
```

mileslucas (approved):

src/methods.jl
siddharthlal25: Sorry about this 😅

giordano:
> Do you mind removing this?

I think it's good to have a reference to the issue, just don't write literally "fixes #..." which will automatically close the issue for the third time. Write "see #..." or anything else, but don't use "fix", "close" or the other special keywords.

Regarding this pull request, it looks overall good, but I feel like there is a lot of code duplication. Maybe you should be using metaprogramming as suggested by Miles?

siddharthlal25:
> Regarding this pull request, it looks overall good, but I feel like there is a lot of code duplication. Maybe you should be using metaprogramming as suggested by Miles?

True, there is quite a lot of code duplication. I was thinking to set up the skeleton for `FITS` interface and then make a separate PR for bundling all things up. Thoughts?

src/methods.jl
```diff
@@ -127,13 +155,23 @@ In-place version of [`flat_correct`](@ref)
 # See Also
 [`flat_correct`](@ref)
 """
-function flat_correct!(frame::AbstractArray, flat_frame::AbstractArray; norm_value = mean(flat_frame))
+function flat_correct!(frame::AbstractArray{T}, flat_frame::AbstractArray; norm_value = mean(flat_frame)) where T
     norm_value <= 0 && error("norm_value must be positive")
-    frame ./= (flat_frame ./ norm_value)
+    @. frame = convert_value(T, frame / (flat_frame / norm_value))
```
mileslucas: no, we shouldn't do this. 

The reason this was introduced for `subtract_overscan`, as we discussed on Zulip, was because we don't yet have an interface for specifying *how* to model the overscan region.

In this case we need to be much more exact, so we can't just blindly round and convert floats to ints, for instance. 

In this case, we _should_ just let the code error if the left hand side (`frame`) does not have a compatible type. Using the `!` version shifts this responsibility to the user!

However, for the non `!` versions, we'll want to convert the frame (and are allowed to!) before passing it on-
```julia
function flat_correct(frame::AbstractArray{T}, flat::AbstractArray{S}, kwargs...) where T,S
    V = float(T, S)
    flat_correct!(V.(frame), flat, kwargs...)
end
```

src/methods.jl
```diff
@@ -127,13 +155,23 @@ In-place version of [`flat_correct`](@ref)
...
     return frame
 end
 
 
+"""
+    flat_correct!(frame::AbstractArray, flat_frame::ImageHDU; norm_value = mean(flat_frame))
+    flat_correct!(frame::AbstractArray, flat_frame::String; hdu = 1, norm_value = mean(flat_frame))
+
+Load a FITS file or HDU for the flat frame before correcting from `frame` in-place.
```
```suggestion
Load a FITS file or HDU for the flat frame before correcting `frame` in-place.
```

src/methods.jl
```diff
@@ -293,7 +349,7 @@ crop(frame, shape; kwargs...) = copy(cropview(frame, shape; kwargs...))
 
 """
     crop(::FITSIO.ImageHDU, shape; force_equal = true)
-    crop(filename, shape; hdu=1, force_equal = true)
+    crop(filename, shape; hdu=1; force_equal = true)
```
```suggestion
    crop(filename, shape; hdu=1, force_equal = true)
```

mileslucas (changes requested):

src/methods.jl
siddharthlal25: True, rounding off can lead to loss of important data! I will make the changes and push the code!

src/methods.jl
siddharthlal25: Missed the semicolon:sweat_smile:

src/methods.jl
mileslucas: I had a small typo, should be
```julia
V = float(promote_type(T, S))
```

src/methods.jl
```diff
@@ -129,11 +157,21 @@ In-place version of [`flat_correct`](@ref)
 """
 function flat_correct!(frame::AbstractArray, flat_frame::AbstractArray; norm_value = mean(flat_frame))
     norm_value <= 0 && error("norm_value must be positive")
-    frame ./= (flat_frame ./ norm_value)
+    frame ./= (flat_frame / norm_value)
```
mileslucas: you shouldn't remove that second dot! Without out it you are missing out on broadcast fusion!

mileslucas (approved):

mileslucas:
> True, there is quite a lot of code duplication. I was thinking to set up the skeleton for FITS interface and then make a separate PR for bundling all things up.

I think it's okay to work your way up to make sure all the tests and expected behavior works out as a learning experience. Once you get more comfortable with metaprogramming, though, you'll see it saves a lot of time and lines.

giordano (approved):

PR #19, opened 2020-06-08
Added FITS interface for `subtract_dark`

### LINKED ISSUE #11: Add interface for FITS files

siddharthlal25:
This PR adds an interface for handling `FITS` file for `subtract_dark` and `subtract_dark!`

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -393,7 +393,7 @@ end
     subtract_dark!(frame::AbstractArray, dark_frame::FITSIO.ImageHDU; data_exposure = 1, dark_exposure = 1)
     subtract_dark!(frame::AbstractArray, dark_frame::String; hdu = 1, data_exposure = 1, dark_exposure = 1)
 
-Load a FITS file or HDU for the dark frame before subtracting from `frame` in-place.
+Load a FITS file or HDU for the dark frame before subtracting from `frame` in-place. If `dark_exposure` is a symbol it will be parsed from the FITS header (case sensitive).
 """
 function subtract_dark!(frame::AbstractArray, dark_frame::ImageHDU; dark_exposure = 1, kwargs...)
 	if dark_exposure isa Symbol
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -19,9 +19,9 @@ convert_value(S, x) = convert(S, x)
 
 # Gets the data from hdu
 function getdata(hdu::ImageHDU)
-	data = read(hdu)
-	d = ndims(data)
-	return permutedims(data, d:-1:1)
+    data = read(hdu)
+    d = ndims(data)
+    return permutedims(data, d:-1:1)
 end
 
 #-------------------------------------------------------------------------------
@@ -109,8 +109,8 @@ subtract_overscan(frame, idxs; kwargs...) = subtract_overscan!(deepcopy(frame), 
 
 
 """
-	subtract_overscan(::FITSIO.ImageHDU, idxs; [dims])
-	subtract_overscan(filename, idxs; hdu=1, [dims])
+    subtract_overscan(::FITSIO.ImageHDU, idxs; [dims])
+    subtract_overscan(filename, idxs; hdu=1, [dims])
 
 Load a FITS file or HDU before subtracting the overscan region. If `idxs` is a symbol it will be read from the FITS header with that key (case sensitive).
 """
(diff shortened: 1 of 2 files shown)


src/methods.jl
```diff
@@ -389,6 +389,22 @@ function subtract_dark!(frame::AbstractArray, dark_frame::AbstractArray; data_ex
 end
 
 
+"""
+    subtract_dark!(frame::AbstractArray, dark_frame::FITSIO.ImageHDU; data_exposure = 1, dark_exposure = 1)
+    subtract_dark!(frame::AbstractArray, dark_frame::String; hdu = 1, data_exposure = 1, dark_exposure = 1)
+
+Load a FITS file or HDU for the dark frame before subtracting from `frame` in-place.
```
```suggestion
Load a FITS file or HDU for the dark frame before subtracting from `frame` in-place. If `dark_exposure` is a symbol it will be parsed from the FITS header (case sensitive).
```

mileslucas (changes requested):

src/methods.jl
```diff
@@ -417,4 +433,39 @@ julia> subtract_dark(frame, dark_frame, data_exposure = 1, dark_exposure = 4)
...
+end
+
+function subtract_dark(frame::AbstractArray, dark_frame::ImageHDU; dark_exposure = 1, kwargs...)
+	if dark_exposure isa Symbol
+		dark_exposure = read_header(frame)[string(dark_exposure)]
+	end
+	return subtract_dark(frame, getdata(dark_frame); dark_exposure = dark_exposure, kwargs...)
+end
+
+subtract_dark(frame::ImageHDU, dark_frame::ImageHDU; kwargs...) = subtract_dark(getdata(frame), dark_frame; kwargs...)
```
mileslucas: You should add a test for 
```julia
subtract_dark(hdu, hdu2, data_exposure=:EXPOSURE)
```
you'll notice it fails

mileslucas (commented):
Looks good- I've noticed all your indentations have turned into tabs instead of spaces. I don't personally care which is used, as long as it's consistent. Currently the files in this repo are a mix of spaces and tabs.

test/methods.jl
```diff
@@ -177,6 +177,35 @@ end
...
+    @test subtract_dark(array_frame, hdu_bias_frame; data_exposure = 2, dark_exposure = 2) == zeros(1059, 1059) # testing Array ImageHDU case
+    @test subtract_dark(hdu_frame, hdu_bias_frame; data_exposure = 2, dark_exposure = 2) == zeros(1059, 1059) # testing ImageHDU ImageHDU case
+    @test subtract_dark(hdu_frame, array_bias_frame; data_exposure = 2, dark_exposure = 2) == zeros(1059, 1059) # testing ImageHDU Array case
+    @test subtract_dark(array_frame, string_bias_frame; data_exposure = 2, dark_exposure = 2) == zeros(1059, 1059) # testing Array String case
+    @test subtract_dark(string_frame, array_bias_frame; data_exposure = 2, dark_exposure = 2) == zeros(1059, 1059) # testing String Array case
+    @test subtract_dark(string_frame, string_bias_frame; data_exposure = 2, dark_exposure = 2) == zeros(1059, 1059) # testsing String String case
+    @test subtract_dark(string_frame, hdu_bias_frame; data_exposure = 2, dark_exposure = 2) == zeros(1059, 1059) # testing String ImageHDU case
+    @test subtract_dark(hdu_frame, string_bias_frame; data_exposure = 2, dark_exposure = 2) == zeros(1059, 1059) # testing ImageHDU String case
+
```
mileslucas, line 200: need to add tests with the hdu which uses `*_exposure=:EXPOSURE`

siddharthlal25:
> Looks good- I've noticed all your indentations have turned into tabs instead of spaces. I don't personally care which is used, as long as it's consistent. Currently the files in this repo are a mix of spaces and tabs.

There is something wrong with my editor, let me figure out how to fix it.
Edit:
Are you referring to inconsistencies in documentation alone or documentation + code?

mileslucas:
I'm referring to the indents in some of the code are now tabs instead of spaces.

siddharthlal25:
> I'm referring to the indents in some of the code are now tabs instead of spaces.

Can you please point me to a few such lines.
PS: I can't find a way to differentiate between tabs and spaces visually in my editor.

siddharthlal25:
I will change all tabs to spaces once the PR gets reviewed by Mosè as well, it can be very tedious to review the PR with all the tabs space changes.
Edit:
The changes were quite small, so I pushed them

giordano (approved):

mileslucas (approved):

PR #20, opened 2020-06-10
Restructured code and tests

### LINKED ISSUE #22: Catch the warnings in the tests
The tests currently issue some warnings like this:
```
┌ Warning: dimension 1 changed from 4 to 5
└ @ CCDReduction ~/.julia/dev/CCDReduction/src/methods.jl:301
┌ Warning: dimension 2 changed from 4 to 5
└ @ CCDReduction ~/.julia/dev/CCDReduction/src/methods.jl:301
┌ Warning: dimension 1 changed from 3 to 4
└ @ CCDReduction ~/.julia/dev/CCDReduction/src/methods.jl:301
┌ Warning: dimension 2 changed from 3 to 4
└ @ CCDReduction ~/.julia/dev/CCDReduction/src/methods.jl:301
┌ Warning: dimension 1 changed from 2 to 3
└ @ CCDReduction ~/.julia/dev/CCDReduction/src/methods.jl:301
┌ Warning: dimension 2 changed from 2 to 3
└ @ CCDReduction ~/.julia/dev/CCDReduction/src/methods.jl:301
```
We should use [`@test_logs`](https://docs.julialang.org/en/v1/stdlib/Test/#Test.@test_logs) to catch them.  This has a two-fold advantage: the warnings aren't shown when running the tests, and we can make sure that the warnings are actually issued as expected.

siddharthlal25:
Switched to `codegen` keeping DRY principle in mind.
Sorry for the messed up diff.


siddharthlal25:
Should I make a separate file `interface.jl` (or `fitsinterface.jl`) for keeping the code of `FITS` interface?
PS: `methods.jl` has become quite lengthy
Edit: 
How about adding `helper.jl`? This would make the code look very neat.

src/methods.jl
```diff
@@ -496,6 +403,105 @@ function subtract_dark(frame::AbstractArray{T}, dark_frame::AbstractArray{S}; kw
...
+for func in (:flat_correct!, :subtract_bias!)
+    @eval $func(frame::AbstractArray, correction::ImageHDU; kwargs...) = $func(frame, getdata(correction); kwargs...)
+    @eval $func(frame::AbstractArray, correction::String; hdu = 1, kwargs...) = $func(frame, FITS(correction)[hdu]; kwargs...)
+end
+
+for func in [:crop, :trim, :subtract_overscan]
+    @eval $func(frame::ImageHDU, args...; kwargs...) = $func(getdata(frame), args...; kwargs...)
+    @eval $func(filename::String, args...; hdu = 1, kwargs...) = $func(FITS(filename)[hdu], args...; kwargs...)
+end
+
```
giordano: I don't think `@eval` is necessary here, as the functions are defined above: don't quote the names of the functions, don't use `@eval`, don' interpolate `func`.

siddharthlal25: I tried something like this:
```
for func in (flat_correct!, subtract_bias!)
    func(frame::AbstractArray, correction::ImageHDU; kwargs...) = func(frame, getdata(correction); kwargs...)
    func(frame::AbstractArray, correction::String; hdu = 1, kwargs...) = func(frame, FITS(correction)[hdu]; kwargs...)
end
```
But now the function does not get defined, it errors in tests.
Edit: 
The error messages are:
```
MethodError: no method matching flat_correct(::String, ::String; norm_value=1)
```

mileslucas: I just tested- that won't work.
```julia
julia> f(x::Number) = 2x
f (generic function with 1 method)

julia> f(2)
4

julia> for func in [f]
           func(s::String) = func(parse(Float64, s))
       end

julia> f("2")
ERROR: MethodError: no method matching f(::String)
```

Edit: yes, I think we have to use `@eval` here.

mileslucas:
> How about adding helper.jl? This would make the code look very neat.

`helper.jl` is pretty ambiguous. How about `fits.jl`.

src/fits.jl
```diff
@@ -0,0 +1,153 @@
...
+            return $func(FITS(frame)[hdus[1]], FITS(correction)[hdus[2]]; kwargs...)
+        end
+    end
+end
+
+for func in (:flat_correct!, :subtract_bias!)
+    @eval $func(frame::AbstractArray, correction::ImageHDU; kwargs...) = $func(frame, getdata(correction); kwargs...)
+    @eval $func(frame::AbstractArray, correction::String; hdu = 1, kwargs...) = $func(frame, FITS(correction)[hdu]; kwargs...)
+end
+
```
mileslucas, line 47: I would add some developer comments here explaining the code gen. For instance, what is different between these for-loops and the above for-loops?

src/fits.jl
```diff
@@ -0,0 +1,153 @@
+# helper functions
+
+# Gets the data from hdu
```
mileslucas: not sure if this is an old or new comment, but I'd specify _why_ this function exists- mention how the data is read transposed from FITSIO.jl

mileslucas (changes requested):

src/fits.jl
```diff
@@ -0,0 +1,153 @@
...
+    trim(::FITSIO.ImageHDU, idxs)
+    trim(filename, idxs; hdu=1)
+
+Load a FITS file or HDU before trimming. If `idxs` is a symbol it will be read from the FITS header with that key (case sensitive).
+"""
+function trim end
+
+
+"""
+    flat_correct(frame, flat_frame; [hdu = 1], norm_value = mean(flat_frame))
```
giordano: Here and also in other places below: I don't think you should use `[...]` around the keyword arguments in the docstrings.  It usually indicates an optional argument, but keyword arguments are always optional.  This is also used in an inconsistent way, only around `hdu`

giordano (commented):
Just a comment about the docstrings

src/fits.jl
siddharthlal25: Here `hdu` is an optional keyword argument, as in at times it is possible that the constructor `flat_frame` will not have `hdu` as a keyword argument. `hdu` is only present in the constructor when `frame`, `flat_frame` or both are passed as strings, otherwise it is not present as a keyword argument. So, I think it is correct to keep `hdu` as `[hdu = 1]`, what's your take on this?
Edit:
For example when `frame` and `flat_frame` both are `FITSIO.ImageHDU`, then `hdu` will not be as a keyword argument.

giordano: Ok, I see what you mean, then it's probably fine

mileslucas: what's awkward here, is that this docstring doesn't match other, similar docstrings. There already exists a docstring for `flat_correct` that defines most of the function. See below how `subtract_overscan` uses a different docstring?
```julia
    subtract_overscan(::FITSIO.ImageHDU, idxs; [dims])
    subtract_overscan(filename, idxs; hdu=1, [dims])
```
why not switch to that for consistency?

mileslucas (approved):

src/fits.jl
siddharthlal25: If we write out all variable types it takes in, it becomes quite clumsy. You suggested to switch for `subtract_bias` as (https://github.com/JuliaAstro/CCDReduction.jl/pull/18#discussion_r436244444) so I switched here as well. Thoughts?

siddharthlal25: Should we switch to those verbose docstrings with all types explicitly mentioned?

mileslucas: I think for now it's nice to show the different docstrings for ImageHDU and strings because they add functionality of choosing the hdu to load and sometimes using  header keys directly with arguments. I think this would work out easier in a separate docstring for now. In the future, we could try to find a way to succinctly incorporate all the documentation in a unified way, but I think this is easier for us and provides the right amount of granularity for users.

PR #23, opened 2020-06-16
Added FITS interface for combine

### LINKED ISSUE #11: Add interface for FITS files

siddharthlal25:


src/fits.jl
```diff
@@ -167,3 +167,17 @@ function subtract_dark!(frame::AbstractArray, dark_frame::ImageHDU; dark_exposur
 end
 
 subtract_dark!(frame::AbstractArray, dark_frame::String; hdu = 1, kwargs...) = subtract_dark!(frame, FITS(dark_frame)[hdu]; kwargs...)
+
+
+"""
+    combine(frames::Vararg{<:FITSIO.ImageHDU}; kwargs...)
+    combine(filename::Vararg{<:String, N}; hdu = ntuple(one, N),  kwargs...)
```
mileslucas: docstrings like this are hard to read, we don't need to copy the method signature exactly (it's easier at that point for users to just read the source code!) Why not something like this-
```suggestion
	combine(frames...; hdu=1, kwargs...)
	
Load multiple FITS files or HDUs before combining. If loading from filenames, the HDU number can be specified with `hdu` as either an integer or a tuple corresponding to each file.
```

src/fits.jl
```diff
@@ -167,3 +167,17 @@ function subtract_dark!(frame::AbstractArray, dark_frame::ImageHDU; dark_exposur
...
+
+Load multiple FITS files or HDUs before combining. Filenames or HDUs can also be passed as vectors for loading before combining.
+"""
+combine(frames::Vararg{<:ImageHDU}; kwargs...) = combine(getdata.(frames)...; kwargs...)
+
+function combine(frames::Vararg{<:String, N}; hdu = ntuple(one, N), kwargs...) where N
+    frames = (FITS(frame)[idx] for (frame, idx) in zip(frames, hdu))
+    combine(frames...; kwargs...)
+end
```
mileslucas: This will fail if I provide `hdu=1` (meaning use the first hdu for all the files). I think this should work-
```suggestion
function combine(frames::Vararg{<:String, N}; hdu = ntuple(one, N), kwargs...) where N
    frames = @. getindex(FITS(frames), hdu)
    combine(frames...; kwargs...)
end
```

src/methods.jl
```diff
@@ -335,7 +335,7 @@ function combine(frames::Vararg{<:AbstractArray}; method = median)
     return reshape(method(stack(frames), dims = dim), shape)
 end
 
-combine(frames::AbstractVector{<:AbstractArray}; method = median) = combine(frames..., method = method)
+combine(frames::AbstractVector; method = median) = combine(frames..., method = method)
```
mileslucas: I think this might be slightly overspecified? It should work fine with just 
```suggestion
combine(frames; method = median) = combine(frames..., method = method)
```

test/fits.jl
```diff
@@ -97,6 +97,24 @@ end
...
+
+    # testing the vector version
+    @test combine(vector_hdu) == combine(vector_frames)
+    @test combine(vector_frames_dir) == combine(vector_frames)
+
+    # testing the varargs version
+    @test combine(M6707HH[1], M6707HH[1], M6707HH[1]) == combine(vector_frames)
+    @test combine(test_file_path_M6707HH, test_file_path_M6707HH, test_file_path_M6707HH) == combine(vector_frames)
+end
+
```
mileslucas, line 120: I would like to see some testing for the heterogeneous hdu (`hdu=1` vs `hdu=(1, 1)` for example)

mileslucas (changes requested):

src/methods.jl
siddharthlal25: Not specifying `AbstractVector` makes this one go wrong:
```
x = [reshape(1:4, (2,2)) for i = 1:10]

@which combine(x)
```
The `combine` goes to:
`combine(frames::AbstractArray...; method) in Main at /home/siddharth/.julia/dev/CCDReduction/src/methods.jl:332` whereas it should have gone to line number 338.

mileslucas: I see. The problem I have with forcing the arg to be an `AbstractVector` is shown by the following:
```julia
x = (reshape(1:4, 2, 2) for i in 1:10)
combine(x) # MethodError
```

Let's leave how it was for now and revisit this later. I'm not too concerned about it right now. This is because the point for using a generator would be to avoid having everything in memory at the same time, but we have to do that anyways to combine them (except for certain cases like using `sum` or `mean` for the combination method).

mileslucas: Actually, I think we can solve this-
all we (might) have to do is change [this signature](https://github.com/JuliaAstro/CCDReduction.jl/blob/c75d5819e1e04d5d9793ec0c44295c14df7a3cf2/src/methods.jl#L331)
to be

```julia
combine(frame::Vararg{<:AbstractArray{<:Number}}, ...)
```
This way, the `[reshape(1:4, (2,2)) for i = 1:10]` example won't dispatch here, because even though it is an `AbstractArray` the element type is not a number.

siddharthlal25: Yes, this worked! :tada:

giordano (approved):
Looks good

src/methods.jl
```diff
@@ -305,10 +305,9 @@ end
 
 
 """
-    combine(frames::Vararg{<:AbstractArray}; method = median)
-    combine(frames::AbstractVector{<:AbstractArray}; method = median)
+    combine(frames::Vararg{<:AbstractArray{<:Number}}; method = median)
```
mileslucas: The docstring here is hard to read! Putting the exact type signatures is distracting, when we can convey the same information easier. Why not put 
`combine(frames...; method=median)` and add the missing signature `combine(frames; method=median)`

mileslucas (approved):

PR #24, opened 2020-06-20
Added `fitscollection`

siddharthlal25:
This PR adds the functionality of searching a location specified by the user and returning all `FITS` files with `hdu` (the ones that are `ImageHDU`) in a data frame.

### DIFF
--- a/src/collection.jl
+++ b/src/collection.jl
@@ -1,13 +1,3 @@
-# We will remove this helper after some versions
-# This is to use endswith in Julia < 1.2
-checkfileending(filename, suffix) = endswith(filename, suffix)
-function checkfileending(filename, suffix::Regex)
-    VERSION ≥ v"1.2" && return endswith(filename, suffix)
-    m = match(suffix, filename)
-    return m !== nothing
-end
-
-#-----------------------------------------------------
 @doc raw"""
     fitscollection(dir; recursive=true, abspath=true, keepext=true, ext=r"fits(\.tar\.gz)?", exclude=nothing, exclude_dir=nothing, exclude_key = ("", "HISTORY"))
 
@@ -68,7 +58,7 @@ function fitscollection(basedir::String;
         end
         for filename in files
             # accept file if .fits or .fits.tar.gz
-            checkfileending(filename, ext) || continue
+            endswith(filename, ext) || continue
             # excluding the files specified by user
             if exclude !== nothing
                 occursin(exclude, filename) && continue
--- a/src/collection.jl
+++ b/src/collection.jl
@@ -1,3 +1,13 @@
+# We will remove this helper after some versions
+# This is to use endswith in Julia < 1.2
+checkfileending(filename, suffix) = endswith(filename, suffix)
+function checkfileending(filename, suffix::Regex)
+    VERSION ≥ v"1.2" && return endswith(filename, suffix)
+    m = match(suffix, filename)
+    return m !== nothing
+end
+
+#-----------------------------------------------------
 @doc raw"""
     fitscollection(dir; recursive=true, abspath=true, keepext=true, ext=r"fits(\.tar\.gz)?", exclude=nothing, exclude_dir=nothing, exclude_key = ("", "HISTORY"))
 
@@ -58,7 +68,7 @@ function fitscollection(basedir::String;
         end
         for filename in files
             # accept file if .fits or .fits.tar.gz
-            endswith(filename, ext) || continue
+            checkfileending(filename, ext) || continue
             # excluding the files specified by user
             if exclude !== nothing
                 occursin(exclude, filename) && continue
--- a/src/collection.jl
+++ b/src/collection.jl
@@ -1,14 +1,18 @@
 # helper function to match extension
 # this will be removed when there is a compat entry of endswith with regex compatibility
 function match_extension(str::String, ext::Regex)
-    r = deepcopy(ext)
-    if !endswith(r.pattern, "\$")
-        r = Regex(r.pattern * "\$", r.compile_options, r.match_options)
+    if VERSION < v"1.2"
+        r = deepcopy(ext)
+        if !endswith(r.pattern, "\$")
+            r = Regex(r.pattern * "\$", r.compile_options, r.match_options)
+        end
+        return occursin(r, str)
+    else
+        endswith(str, ext)
     end
-    return occursin(r, str)
 end
 
-match_extension(str::String, ext::String) = endswith(str, ext)
+match_extension(str::String, ext) = endswith(str, ext)
 
 #------------------------------------------------------------------------------------------------
 @doc raw"""
@@ -39,16 +43,16 @@ For more information about the file matching and path deconstruction, see the ex
 
 ## Parts of a path
 
-Here is an example of a file path and how it would be parsed
+Let's look at some file paths starting from `"/data"`. Here are examples of how they would be parsed
 
 ```
-     root         dir   base   ext
-[---------------][----][-----][----]
-\Users\miles\data\test\tek0001.fits
+ root  dir   base   ext
+[----][---][------][---]
+/data/test/tek0001.fits
 
-     root          dir     base    ext
-[---------------][--------][-----][----]
-\Users\miles\data\test\sci\tek0001.fits
+ root    dir     base   ext
+[----][-------][------][---]
+/data/test/sci/tek0001.fits
 ```
 
 If `keepext` is `true`, `name=base * ext`, otherwise it is just `base`. If `abspath` is `true`, the path will be `root * dir * base * ext`, otherwise it will be `dir * base * ext`. These options allow flexility in creating a table that can be easily saved and loaded to avoid having to manually filter files. Especially consider how `abspath` can allow keeping tables that will transfer easily between computers or between data sources with common structures.
@@ -88,9 +92,7 @@ function fitscollection(basedir::String;
 
                 # filtering out comment columns
                 _keys = filter(k -> k ∉ exclude_key, keys(header_data))
-                _values = map(_keys) do k
-                    header_data[k]
-                end
+                _values = (header_data[k] for k in _keys)
                 push!(df, (path = path, name = name, hdu = index, zip(Symbol.(_keys), _values)...); cols = :union)
             end
             close(fits_data)
--- a/src/collection.jl
+++ b/src/collection.jl
@@ -1,0 +1,71 @@
+"""
+    filecollection(dir; walk=true, abspath=true, keepext=true, ext=r"fits(\.tar\.gz)?", exclude=nothing, dir_exclude=nothing)
+
+Walk through `dir` collecting FITS files, scanning their headers, and culminating into a `DataFrame` that can be used with the [generators](@ref gens) for iterating over many files and processing them. If `walk` is false, no subdirectories will be walked through.
+
+The table returned will contain the path to the file, the name of the file, and index of the corresponding hdu, and each FITS header column and value. If two FITS files have distinct columns, they will both appear in the table with `missing` in the appropriate rows.
+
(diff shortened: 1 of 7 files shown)


src/collection.jl
```diff
@@ -0,0 +1,71 @@
+raw"""
+    filecollection(dir; walk=true, abspath=true, keepext=true, ext=r"fits(\.tar\.gz)?", exclude=nothing, dir_exclude=nothing)
+
+Walk through `dir` collecting FITS files, scanning their headers, and culminating into a `DataFrame` that can be used with the [generators](@ref gens) for iterating over many files and processing them. If `walk` is false, no subdirectories will be walked through.
```
giordano: Maybe call the argument `recursive` instead of `walk`?  This is consistent with the name of argument of `rm` for example, which has the same meaning

src/collection.jl
```diff
@@ -0,0 +1,71 @@
...
+
+The table returned will contain the path to the file, the name of the file, and index of the corresponding hdu, and each FITS header column and value. If two FITS files have distinct columns, they will both appear in the table with `missing` in the appropriate rows.
+
+If `abspath` is true, the path in the table will be absolute compared to `dir`. If `keepext` is true, the name in the table will include the file extension, given by `ext`. `ext` will be used with `endswith` to filter for fits files compatible with `FITSIO.FITS`. `exclude` is a pattern that can be used with `occursin` to exclude certain filenames. For example, to exclude any files containing "sky",
```
giordano: > the path in the table will be absolute compared to `dir`.

I find this very confusing.  An absolute path is absolute by definition (i.e., is not relative to anything), what do you by "compare to `dir`"?

src/collection.jl
siddharthlal25: Sure!

src/collection.jl
siddharthlal25: The docs are very rough as of now, I think it should have been: "If `abspath` is true, the path in the table will be absolute."

siddharthlal25:
Any standard way of testing the functions with large outputs? @mileslucas @giordano

src/collection.jl
```diff
@@ -39,10 +39,12 @@ Here is an example of a file path and how it would be parsed
 
 If `keepext` is `true`, `name=base * ext`, otherwise it is just `base`. If `abspath` is `true`, the path will be `root * dir * base * ext`, otherwise it will be `dir * base * ext`. These options allow flexility in creating a table that can be easily saved and loaded to avoid having to manually filter files. Especially consider how `abspath` can allow keeping tables that will transfer easily between computers or between data sources with common structures.
 """
-function fitscollection(basedir::String; abspath = true, keepext = true, ext = r"fits(\.tar\.gz)?"i, exclude = nothing, dir_exclude = nothing, recursive = true)
+function fitscollection(basedir::String; recursive = true, abspath = true, keepext = true, ext = r"fits(\.tar\.gz)?"i, exclude = nothing, dir_exclude = nothing)
```
mileslucas: feel free to break this up into multiple lines
```suggestion
function fitscollection(basedir::String;
		recursive = true,
		abspath = true,
		keepext = true,
		ext = r"fits(\.tar\.gz)?"i,
		exclude = nothing,
		dir_exclude = nothing)
```

src/collection.jl
```diff
@@ -39,7 +39,13 @@ Here is an example of a file path and how it would be parsed
...
 If `keepext` is `true`, `name=base * ext`, otherwise it is just `base`. If `abspath` is `true`, the path will be `root * dir * base * ext`, otherwise it will be `dir * base * ext`. These options allow flexility in creating a table that can be easily saved and loaded to avoid having to manually filter files. Especially consider how `abspath` can allow keeping tables that will transfer easily between computers or between data sources with common structures.
 """
-function fitscollection(basedir::String; recursive = true, abspath = true, keepext = true, ext = r"fits(\.tar\.gz)?"i, exclude = nothing, dir_exclude = nothing)
+function fitscollection(basedir::String;
+		recursive = true,
+		abspath = true,
+		keepext = true,
+		ext = r"fits(\.tar\.gz)?"i,
+		exclude = nothing,
+		dir_exclude = nothing)
```
mileslucas: Github may have automatically done this, but these tabs need to be turned to spaces

siddharthlal25: I will write the tests, switch tabs to spaces and push the code by today!

siddharthlal25:
This PR is currently on hold, waiting for (https://github.com/JuliaLang/Compat.jl/issues/707) issue to be resolved.

giordano:
Why is `endswith` necessary?  Can't you just match a regular expression?

siddharthlal25:
Yes we can do that, but then the user has to type in `$` at the end of regex (to signify it matches from the end)
PS: I don't think a user would prefer putting `$` at the end

mileslucas:
> Can't you just match a regular expression?

I played around with some workarounds, but the issue is that we have to match the end of the line. A contrived example of a failure would look like this-
```julia
file = "fits_test.fi"
# this is the default suffix, and doesn't require user to explicitly write the `.` delimiter or the line end `r"$"`.
suffix = r"fits((\.tar)?\.gz)?"i 
occursin(suffix, file) # true, matches `fits` in `fits_test`
```
so to elude the `endswith` compat issues, we _could_ write it in a way that forces a regex expression to have the line end `r"$"` in it-
```julia
# in some code that matches only if suffix isa Regex and VERSION < v"1.2"
if !endswith(suffix.pattern "\$")
	suffix.pattern *= "\$"
end
return occursin(suffix, filename)
```

giordano:
> Yes we can do that, but then the user has to type in $ at the end of regex (to signify it matches from the end)

Why do they need to?  You can append it

giordano:
```julia
julia> function match_extension(ext::Regex, str::String)
           r = deepcopy(ext)
           if !endswith(r.pattern, "\$")
               r = Regex(r.pattern * "\$", r.compile_options, r.match_options)
           end
           return occursin(r, str)
       end
match_extension (generic function with 1 method)

julia> match_extension(r"fits((\.tar)?\.gz)?"i, "fits_test.fi")
false
```

Fiddling too much with internals?  I wouldn't hold my breath until someone else implements a function in `Compat`: you either go ahead and implement it yourself or just work around the compatibility limitation here.

siddharthlal25:
Cool! Let me add some more documentation and this would look good!

siddharthlal25:
`fitscollection` vs `filecollection`, thoughts?

giordano (approved):
As I wrote in the comments, `match_extension` might be a tad too internal-y, apart from that this looks good to me

src/collection.jl
```diff
@@ -0,0 +1,100 @@
+# helper function to match extension
+# this will be removed when there is a compat entry of endswith with regex compatibility
+function match_extension(str::String, ext::Regex)
+    r = deepcopy(ext)
+    if !endswith(r.pattern, "\$")
+        r = Regex(r.pattern * "\$", r.compile_options, r.match_options)
+    end
```
mileslucas: do we need to deepcopy if you're creating a new copy of Regex? Or vice-versa? I feel like we could get away with
```julia
r = deepcopy(ext)
if !endswith(r.pattern, "\$")
    r.pattern. *= "\$"
end
```

Also, wouldn't it be worth it to wrap the whole thing in an `if` block for `VERSION < v"1.2"`, at least for now?

src/collection.jl
```diff
@@ -0,0 +1,100 @@
...
+
+```
+     root         dir   base   ext
+[---------------][----][-----][----]
+\Users\miles\data\test\tek0001.fits
+
+     root          dir     base    ext
+[---------------][--------][-----][----]
+\Users\miles\data\test\sci\tek0001.fits
+```
```
mileslucas: I think this can be cleaned up a little-

```julia
Let's look at some file paths starting from `"/data"`. Here are examples of how they would be parsed

\```
 root  dir   base   ext
[----][---][------][---]
/data/test/tek0001.fits

 root    dir     base   ext
[----][-------][------][---]
/data/test/sci/tek0001.fits
\```
```

src/collection.jl
```diff
@@ -0,0 +1,100 @@
...
+                hdu isa ImageHDU || continue
+                header_data = read_header(hdu)
+                path = abspath ? Base.abspath(location) : location
+                name = keepext ? filename : first(split(filename, ext))[1:end-1]
+
+                # filtering out comment columns
+                _keys = filter(k -> k ∉ exclude_key, keys(header_data))
+                _values = map(_keys) do k
+                    header_data[k]
+                end
```
mileslucas: this is a style thing more than anything, but I don't feel like we need the `do` syntax here. I would have simply done
```julia
_values = (header_data[k] for k in _keys)
```

test/collection.jl
```diff
@@ -0,0 +1,45 @@
...
+
+    # the two fits file used for testing have identical header keys
+    df = fitscollection(dir)
+    relevant_keys = filter(i -> i ∉ ("", "HISTORY"), keys(read_header(M6707HH[1])))
+    relevant_values_M6707HH = map(relevant_keys) do key
+        read_header(M6707HH[1])[key]
+    end
+    relevant_values_M35070V = map(relevant_keys) do key
+        read_header(M35070V[1])[key]
+    end
```
mileslucas: I would be wary of calling `read_header` so many times- every single call is a read from disk, then getting one value from it. Compared to loading the headers once and iterating like so.

test/collection.jl
```diff
@@ -0,0 +1,45 @@
...
+    # testing rows columns
+    # an additional 3 columns are present because of varibles path, name and hdu
+    @test size(df) == (2, length(relevant_keys) + 3)
+
+    # testing each element in rows and columns
+    @test copy(df[1, :]) == (path = joinpath(dir, "M35070V.fits"), name = "M35070V.fits", hdu = 1, zip(Symbol.(relevant_keys), relevant_values_M35070V)...)
+    @test copy(df[2, :]) == (path = joinpath(dir, "M6707HH.fits"), name = "M6707HH.fits", hdu = 1, zip(Symbol.(relevant_keys), relevant_values_M6707HH)...)
```
mileslucas: Style:  might be worth constructing the tuple with line-breaks. instead of one long line

src/collection.jl
```diff
@@ -0,0 +1,100 @@
...
+# this will be removed when there is a compat entry of endswith with regex compatibility
+function match_extension(str::String, ext::Regex)
+    r = deepcopy(ext)
+    if !endswith(r.pattern, "\$")
+        r = Regex(r.pattern * "\$", r.compile_options, r.match_options)
+    end
+    return occursin(r, str)
+end
+
+match_extension(str::String, ext::String) = endswith(str, ext)
```
```suggestion
match_extension(str::String, ext) = endswith(str, ext)
```

I would avoid specializing the extension in this version

mileslucas (changes requested):

src/collection.jl
siddharthlal25: Ya, an if-else statement would be good!

siddharthlal25: Some tests fail for this version of `match_extension`
Edit:
This failed ` @test !match_extension("abcd.fits.tar.gz", r"fits(\.gz)?"i)`

giordano: Simply changing `pattern` doesn't change the compiled regexp, that's why I used the constructor.

siddharthlal25: Can you explain a bit about what exactly is `compile_options` and `match_options`?

giordano: No because I don't know _exactly_ what they are :slightly_smiling_face:  I assume `match_options` encodes the options like case-sensitivity etc.

mileslucas: ah makes, sense! In that case, we can still elude the deepcopy, I believe.

giordano: Since `Regex` is a mutable struct I wanted to avoid mutating the input object.

mileslucas (approved):

PR #26, opened 2020-06-28
Bump julia to v1.3

siddharthlal25:
This is being done to use the package `ResumableFunctions.jl` for creating generators.

### DIFF
--- a/src/collection.jl
+++ b/src/collection.jl
@@ -1,3 +1,15 @@
+# helper function
+# parses the name and returns it with or without extension
+parse_name(filename, ext::AbstractString, ::Val{false}) = first(rsplit(filename, ext, limit=2))
+
+function parse_name(filename, ext::Regex, ::Val{false})
+	idxs = findall(ext, filename)
+    return filename[1:first(last(idxs)) - 1]
+end
+
+parse_name(filename, ext, ::Val{true}) = filename
+
+#---------------------------------------------------------------------------------------
 @doc raw"""
     fitscollection(dir; recursive=true, abspath=true, keepext=true, ext=r"fits(\.tar\.gz)?", exclude=nothing, exclude_dir=nothing, exclude_key = ("", "HISTORY"))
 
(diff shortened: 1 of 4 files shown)


mileslucas (approved):

src/collection.jl
```diff
@@ -88,7 +71,7 @@ function fitscollection(basedir::String;
                 hdu isa ImageHDU || continue
                 header_data = read_header(hdu)
                 path = abspath ? Base.abspath(location) : location
-                name = keepext ? filename : first(split(filename, ext))[1:end-1]
+                name = keepext ? filename : first(split(filename, "." * ext))
```
giordano: Does this work if the file has a funny name like `foo.fits.fits`?

siddharthlal25: Doing this gives us only `foo`, I hope we don't get something strange like this. :joy:

giordano: Well, don't expect users to choose sensible names :wink:

giordano: I didn't try, but doesn't something like
```suggestion
                name = keepext ? filename : join(split(filename, "." * ext)[1:end-1])
```
work?

siddharthlal25: What would you suggest to tackle this case?

siddharthlal25: Tried something like this:
```
ext = r"fits(\.tar\.gz)?"i
filename = "foo.fits.fits"

join(split(filename, "." * ext)[1:end-1])
```
This gives `foo` only

Edit:
```
split(filename, "." * ext)
```
gives: 
```
Vector{SubString{String}} with 3 elements
"foo"
""
""
```

mileslucas:
```julia
parse_name(filename, ext::AbstractString, ::Val{false}) = first(rsplit(filename, ext, limit=2))

function parse_name(filename, ext::Regex, ::Val{false})
	idxs = findall(ext, filename)
    return filename[1:first(last(idxs)) - 1]
end

# within the parsing function
name = parse_name(filename, "." * ext, Val(keepext))
```

siddharthlal25: Should we modify the `ext` by adding `$` at the end and then just call `first(split(filename, "." ext))`?

siddharthlal25: >     return filename[1:first(last(idxs)) - 1]
Shouldn't this be `[1:first(last(idxs)) - 2]`?

siddharthlal25: No, it should be 1 only.

giordano (approved):

PR #27, opened 2020-06-28
Generators for iterating through file collection

### LINKED ISSUE #25: Generators for iterating through filecollection
Here is a rough start to defining special iterators for going through the dataframes from `filecollection`

## Goal

After using `filecollection` something I might want to do is create a master bias frame. I can do this easily by filtering the collection for the biases (OBJECT="")
```julia
full_coll = filecollection(...)
bias_coll = filter(r -> r.OBJECT == "", full_coll)
```
now, I want to be able to use the methods like `combine` using this dataframe! Here is a simple example-

```julia
bias_arr = map(images(bias_coll)) do img
    # img isa ImageHDU
	tmp = subtract_overscan(img, :BIASSEC)
	trim!(tmp, :BIASSEC)
end
master_bias = combine(bias_arr)
```

## Methods
There are three methods I want targeted

1. Iterating and returning the filename
2. Iterating and returning an ImageHDU
3. Iterating and returning an Array

These may seem simple right now, but I have lots of ideas for future improvements that will be based off this interface. In terms of implementation, I want to use `Task`s to create generators [this blog post](http://slendermeans.org/julia-iterators.html) shows a good example using a Task as a generator, and the Julia docs explain the functions as well. 

The reason I want to use Tasks is because we want to be able to modify certain data outside of the iteration- for instance, you'll want to do 
```julia
hdu = FITS(row.path)[row.hdu]
try # not sure if this is exact
    produce(hdu)
finally
	close(hdu)
end
```
to make sure the fits files close properly.

siddharthlal25:
This PR adds basic generators interface for iterating on file collection. Related issue: #25

### DIFF
--- a/src/CCDReduction.jl
+++ b/src/CCDReduction.jl
@@ -20,7 +20,8 @@ export subtract_bias,
        subtract_dark,
        subtract_dark!,
        fitscollection,
-       array
+       array,
+       filename
 
 include("methods.jl")
 include("fits.jl")
--- a/src/CCDReduction.jl
+++ b/src/CCDReduction.jl
@@ -4,6 +4,7 @@ using Statistics
 using LazyStack
 using FITSIO
 using DataFrames: DataFrame
+using ResumableFunctions
 
 export subtract_bias,
        subtract_bias!,
@@ -18,7 +19,8 @@ export subtract_bias,
        combine,
        subtract_dark,
        subtract_dark!,
-       fitscollection
+       fitscollection,
+       array
 
 include("methods.jl")
 include("fits.jl")
--- a/src/CCDReduction.jl
+++ b/src/CCDReduction.jl
@@ -20,8 +20,8 @@ export subtract_bias,
        subtract_dark,
        subtract_dark!,
        fitscollection,
-       array,
-       filename,
+       arrays,
+       filenames,
        images
 
 include("methods.jl")
--- a/src/CCDReduction.jl
+++ b/src/CCDReduction.jl
@@ -21,7 +21,8 @@ export subtract_bias,
        subtract_dark!,
        fitscollection,
        array,
-       filename
+       filename,
+       images
 
 include("methods.jl")
 include("fits.jl")
--- a/src/collection.jl
+++ b/src/collection.jl
@@ -3,7 +3,7 @@
 parse_name(filename, ext::AbstractString, ::Val{false}) = first(rsplit(filename, ext, limit=2))
 
 function parse_name(filename, ext::Regex, ::Val{false})
-	idxs = findall(ext, filename)
+    idxs = findall(ext, filename)
     return filename[1:first(last(idxs)) - 1]
 end
 
@@ -120,7 +120,7 @@ end
 
 # generator for ImageHDU specified by data frame (i.e. path of file, hdu etc.)
 @resumable function images(df::DataFrame)
-	for row in eachrow(df)
-		@yield FITS(row.path)[row.hdu]
-	end
+    for row in eachrow(df)
+        @yield FITS(row.path)[row.hdu]
+    end
 end
--- a/src/collection.jl
+++ b/src/collection.jl
@@ -98,7 +98,7 @@ end
 
 
 # generator for image arrays specified by data frames (i.e. path of file, hdu etc.)
-@resumable function array(df::DataFrame)
+@resumable function arrays(df::DataFrame)
     for row in eachrow(df)
         fh = FITS(row.path)
         try
@@ -111,7 +111,7 @@ end
 
 
 # generator for filenames specified by data frame (i.e. path of file, hdu etc.)
-@resumable function filename(df::DataFrame)
+@resumable function filenames(df::DataFrame)
     for row in eachrow(df)
         @yield row.path
     end
--- a/src/collection.jl
+++ b/src/collection.jl
@@ -97,6 +97,13 @@ function fitscollection(basedir::String;
 end
 
 
+"""
+    arrays(df::DataFrame)
+
+Generator for arrays of images of entries in data frame.
+"""
+function arrays end
+
 # generator for image arrays specified by data frames (i.e. path of file, hdu etc.)
 @resumable function arrays(df::DataFrame)
     for row in eachrow(df)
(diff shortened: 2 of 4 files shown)


siddharthlal25:
How to document `@resumable` functions? When I added a block for documentation, it throws an error: 
```
cannot document the following expression:

#= /home/siddharth/.julia/dev/CCDReduction.jl/src/collection.jl:105 =# @resumable function array(df::DataFrame)
```

test/collection.jl
```diff
@@ -45,6 +45,39 @@ using CCDReduction: parse_name
...
+@testset "image-generators" begin
+    # setting initial data
+    dir = joinpath(@__DIR__, "data")
+    df = fitscollection(dir)
+    arr1 = map(x->x, images(df))
+    arr2 = map(eachrow(df)) do row
+        FITS(row.path)[row.hdu]
+    end
+    @test arr1 == arr2
+end
```
siddharthlal25: I can't understand why this test fails, in my machine it shows the following stack:
```
image-generators: Test Failed at /home/siddharth/.julia/dev/CCDReduction.jl/test/collection.jl:78
  Expression: arr1 == arr2
   Evaluated: ImageHDU[File: /home/siddharth/.julia/dev/CCDReduction.jl/test/data/M35070V.fits
HDU: 1
Type: Image
Datatype: Int16
Datasize: (1059, 1059), File: /home/siddharth/.julia/dev/CCDReduction.jl/test/data/M6707HH.fits
HDU: 1
Type: Image
Datatype: Int16
Datasize: (1059, 1059)] == ImageHDU[File: /home/siddharth/.julia/dev/CCDReduction.jl/test/data/M35070V.fits
HDU: 1
Type: Image
Datatype: Int16
Datasize: (1059, 1059), File: /home/siddharth/.julia/dev/CCDReduction.jl/test/data/M6707HH.fits
HDU: 1
Type: Image
Datatype: Int16
Datasize: (1059, 1059)]
```
PS: I think the left side and right side are equal.

src/collection.jl
```diff
@@ -95,3 +95,32 @@ function fitscollection(basedir::String;
...
+
+
+# generator for image arrays specified by data frames (i.e. path of file, hdu etc.)
+@resumable function array(df::DataFrame)
+    for row in eachrow(df)
+        fh = FITS(row.path)
+        try
+            @yield getdata(fh[row.hdu])
+        finally
+            close(fh)
```
giordano: Why not
```suggestion
        FITS(row.path) do fh
            @yield getdata(fh[row.hdu])
```
?

src/collection.jl
```diff
@@ -95,3 +95,32 @@ function fitscollection(basedir::String;
...
+        @yield row.path
+    end
+end
+
+
+# generator for ImageHDU specified by data frame (i.e. path of file, hdu etc.)
+@resumable function images(df::DataFrame)
+	for row in eachrow(df)
+		@yield FITS(row.path)[row.hdu]
+	end
```
giordano: Indentation is different from the previous function

test/collection.jl
giordano: `ImageHDU` is a mutable struct

src/collection.jl
siddharthlal25: This one throws an error:
```
syntax: label "_STATE_1" referenced but not defined
```

giordano: Looks like a bug in `ResumableFunctions.jl`: https://github.com/BenLauwens/ResumableFunctions.jl/issues/28

src/CCDReduction.jl
```diff
@@ -18,7 +19,10 @@ export subtract_bias,
        combine,
        subtract_dark,
        subtract_dark!,
-       fitscollection
+       fitscollection,
+       array,
+       filename,
+       images
```
mileslucas: we should use plural for all of these functions- `images` is an iterator of images, so we use plural. `array` is an iterator of arrays, yet it's being labeled `array`.

On top of that, `array` and `filename` as function names is just asking for confusion/name conflicts. `arrays` and `filenames` is a little less likely.

test/collection.jl
```diff
@@ -45,6 +45,48 @@ using CCDReduction: parse_name
     @test size(df) == (0, 0)
 end
 
+@testset "array-generators" begin
+    # setting initial data
+    dir = joinpath(@__DIR__, "data")
+    df = fitscollection(dir)
+    arr1 = map(x -> x, array(df))
```
mileslucas:
```suggestion
    arr1 = array(df) |> collect
```

test/collection.jl
```diff
@@ -45,6 +45,48 @@ using CCDReduction: parse_name
...
+        getdata(FITS(row.path)[row.hdu])
+    end
+    @test arr1 == arr2
+end
+
+@testset "filename-generators" begin
+    # setting initial data
+    dir = joinpath(@__DIR__, "data")
+    df = fitscollection(dir)
+    arr1 = map(x -> x, filename(df))
```
mileslucas: similar to above, try just using `collect`

mileslucas (changes requested):
the core functionality looks good, needs docs and some changes first.

giordano (approved):
Looks good :+1:

src/collection.jl
```diff
@@ -95,3 +95,53 @@ function fitscollection(basedir::String;
...
+
+# generator for image arrays specified by data frames (i.e. path of file, hdu etc.)
+@resumable function arrays(df::DataFrame)
+    for row in eachrow(df)
+        fh = FITS(row.path)
+        try
+            @yield getdata(fh[row.hdu])
+        finally
+            close(fh)
+        end
```
mileslucas: Does this work? The more file IO safeguards we can alleviate to other libraries, the better.
```suggestion
        FITS(row.path) do fh
            @yield getdata(fh[row.hdu])
        end
```

If this doesn't work, I'd vote to mimick the other two functions and skip the `try` block (just for consistency).

mileslucas (approved):
This looks good for the most part!

src/collection.jl
siddharthlal25: This won't work, this is a bug from `ResumableFunctions.jl`.

Issue #28, opened 2020-06-30
automatic copying within generators


mileslucas:
A very common workflow for reducing data includes copying files after each operation to have a separate copy for each step of the reduction.

For example, I have 10 bias frames and I want to overscan subtract and trim them before saving to a new location. For each file, I want the new name to be `<name>_oc_trim.<ext>`. Or perhaps I want them to be saved in a new folder `oc_trim/<name>.<ext>` (or maybe both!).

We can automate this process using the generators from #27 with some new keyword arguments

- `save` - save the files using the options below.
- `save_folder` - new folder to save in
- `save_prefix` - add to name like `<prefix>_<name>.<ext>`
- `save_suffix` - add to end of name like `<name>_<suffix>.<ext>`
- `save_delim="_"` - the thing to put before/after the prefix or suffix

The general idea is that if any of `save_folder`, `save_prefix`, `save_suffix` is given then the files are copied and saved (ie `save=true`), otherwise they are not (ie `save=false`). The user can manually write `save=false`, for example, to get filenames using the mechanics provided without actually copying data.

Now, in terms of implementation I don't have a perfect idea yet, but I would generally want the users to be able to avoid manually calling any `write` functions. In other words, if I set these keywords appropriately, there should be nothing different in my for-loop compared to a version that doesn't save. To do this, you'll have to find a way using ResumableFunctions to get a "return value". We can discuss further ideas to accomplish this.

PR #30, opened 2020-07-07
Added save compatible version of `filenames`, `arrays` and `images` function

siddharthlal25:
One of the most important functions of this package!
Related issue: #28


src/collection.jl
```diff
@@ -142,3 +152,43 @@ function images end
         @yield FITS(row.path)[row.hdu]
     end
 end
+
+
+@doc raw"""
```
giordano: Out of curiosity, why using `@doc` and the raw string?

src/collection.jl
```diff
@@ -9,6 +9,16 @@ end
 
 parse_name(filename, ext, ::Val{true}) = filename
 
+
+# transposes the data and saves it, the reason is same as that for getdata
+function setdata(file_path, data)
```
giordano: Not sure the name of the function is super clear about what it does?  Probably a docstring would also be in order here

src/collection.jl
```diff
@@ -142,3 +152,43 @@ function images end
         @yield FITS(row.path)[row.hdu]
     end
 end
+
+
+@doc raw"""
+    process(f, df::DataFrame; path = nothing, save_prefix = nothing, save_suffix = nothing, save_delim = "_", ext = r"fits(\.tar\.gz)?"i)
+
+This is a function to apply multiple function `f` on all elements of data frame and then save it.
```
giordano: "multiple"?  It's only one

src/collection.jl
siddharthlal25: Because of this `r"fits(\.tar\.gz)?"i` in documentation.

giordano:
```julia
julia> """
       r"fits(\\.tar\\.gz)?"i
       """
"r\"fits(\\.tar\\.gz)?\"i\n"
```

src/collection.jl
siddharthlal25: We can apply two functions, no?
```
function proc(f, x)
    println(f(x))
end

f1(x) = x + 2
f2(x) = x^2

proc(2) do x
    y = f1(x)
    z = f2(y)
end # outputs 16
```
As in one after the other. (e.g. `subtract_overscan` followed by `trim`)

giordano: That's confusing: the entire body of the `do` block constitutes a single anonymous function, inside which one can do whatever they want.

siddharthlal25: Hmm, true! It's only a single function. I will make the changes in documentation.

src/collection.jl
siddharthlal25: Should we expose this function to the user?

mileslucas (commented):
My ideal interface would have three versions of this, `images`, `arrays`, `filenames`, to match the three functions we currently have that support iteration. This gives a consistent interface across the API- I can go from
```julia
for img in images(coll)
	# ...
end
```
to this:
```julia
images(coll) do img
    # ...
end
```

and get the same functionality, plus the saving _if I want it_.

src/collection.jl
```diff
@@ -9,6 +9,16 @@ end
...
+    d = ndims(data)
+    transposed_data = permutedims(data, d:-1:1)
+    FITS(file_path, "w") do fh
+        write(fh, transposed_data)
+    end
+end
+
 #---------------------------------------------------------------------------------------
 @doc raw"""
     fitscollection(dir; recursive=true, abspath=true, keepext=true, ext=r"fits(\.tar\.gz)?", exclude=nothing, exclude_dir=nothing, exclude_key = ("", "HISTORY"))
```
mileslucas: I'm not sure how this actually looks in the docs, but I imagine it renders poorly with a horizontal scroll bar. If that is the case, I would strongly recommend breaking up the kwargs with newlines

src/collection.jl
```diff
@@ -142,3 +152,94 @@ function images end
...
+        # if path is not nothing then we save
+        if !(path isa Nothing)
+            make_file(processed_value, x.name, path, save_prefix, save_suffix, save_delim, ext)
+        end
+    end
+    return final_value
+end
+
+# utility function to generate file name and then save the given data
+function make_file(data, filename, save_location, save_prefix, save_suffix, save_delim, ext)
```
mileslucas: I would separate the file-name functionality from the saving functionality. This makes documentation and unit-testing much easier, and the only difference in your iterator calls is an extra line, like
```julia
# previous
make_file(...)
# new
save_path = generate_filename(filename, opts...)
write_fits(save_path, data)
```

src/collection.jl
```diff
@@ -142,3 +152,94 @@ function images end
...
+
+
+"""
+    images(f, df::DataFrame; path = nothing, save_prefix = nothing, save_suffix = nothing, save_delim = "_", ext = r"fits(\\.tar\\.gz)?"i)
+
+Applies function `f` on all ImageHDUs present in data frame and saves it in FITS file.
+
+If `path = nothing`, then save functionality does not execute. It returns an array of arrays which contains final returned values of function.
+A suffix and prefix can be added to filename of newly created files by modifying `save_suffix` and `save_prefix`, `save_delim` is used as delimiter.
+`ext` is the extension of files to be taken into consideration for applying function, by default it is set to `r"fits(\\.tar\\.gz)?"i`.
```
mileslucas: We will definitely want to show some examples with this, I'm not sure whether it makes more sense to copy-paste three nearly identical examples in the docstrings or whether to relagate that to the online documentation. Perhaps both?

Another way to get around this is to put all of the filename-changing stuff into the docstring for `make_file` and then show the `CCDReduction.make_file` docstring in the docs and reference it. This removes the repitition and keeps all the relevant info in docstrings, but at the cost of having to search around a little more to get the information you need.

src/collection.jl
```diff
@@ -142,3 +152,94 @@ function images end
...
+"""
+function filenames(f, df::DataFrame; path = nothing, save_prefix = nothing, save_suffix = nothing, save_delim = "_", ext = r"fits(\.tar\.gz)?"i)
+    final_value = Vector{Array}(undef, first(size(df)))
+    for (i,x) in enumerate(eachrow(df))
+        processed_value = f(x.path)
+        final_value[i] = processed_value
+        # if path is not nothing then we save
+        if !(path isa Nothing)
+            make_file(processed_value, x.name, path, save_prefix, save_suffix, save_delim, ext)
+        end
```
mileslucas: style, but on 1.1+ you can use `isnothing` like `!isnothing(path)`. Functionally equivalent, but your choice of what you like.

src/collection.jl
mileslucas: `setdata` is a natural opposite to `getdata`, but unfortunately is not really accurate for what we're doing (creating a new file and writing data to it).

I would recommend a name closer to `write_data` or something that takes a path and data. Maybe adding an overwrite option?

I wouldn't expose this by default (same as `getdata`), but I would keep it well-documented. This is a function I can see being used for convenience, but I don't want it to be part of the public API because it is just some small functionality that we ought to be able to change when/how we want.

src/collection.jl
```diff
@@ -142,3 +152,94 @@ function images end
...
+    for (i,x) in enumerate(eachrow(df))
+        fh = FITS(x.path)
+        processed_value = f(fh[x.hdu])
+        close(fh)
+        final_value[i] = processed_value
+        # if path is not nothing then we save
+        if !(path isa Nothing)
+            make_file(processed_value, x.name, path, save_prefix, save_suffix, save_delim, ext)
+        end
+    end
```
mileslucas: I'm not a big fan of the pre-allocate (and guess type) and push inside a for-loop design, but if it works it works. For something like this I would tend to gravitate towards
```julia
processed_images = map(eachrow(df)) do row
    # do your stuff
    return processed_image
end
```

src/collection.jl
```diff
@@ -142,3 +152,94 @@ function images end
...
+    return final_value
```
mileslucas: Keep an eye out for variable names- `final_value` sounds like I should be getting some single scalar out, but this is actually a vector of matrices! In my example above I used `processed_images` which is clearer (at least to me).

mileslucas (changes requested):
Okay, skimming over the three versions, I think we should have a little more granularity in the functionality, like I mention with splitting the `make_file` function.

In addition, the current mechanics are kind of hard to understand- if `path` is not nothing we will save files. Where will you save files? Is this path supposed to be the load path, too? What is being saved? Can I do a dry-run without actually saving? 

Some of these just require some documentation, but I really think we need to re-evaluate the functionality of the `path` argument in its current state and either change the name, add more documentation, or split its functionality into two parts (eg `path` is just the save path, which defaults to same folder as the file and the save mechanics can be manually enabled/disabled with `save`).

I left some comments about using `map` style instead of the for-loop, in addition I want you to think about how we can use the functions that already exist to greatly simplify our API. Consider what the current `images` function can do- iterate over the dataframe with some options and provide each `ImageHDU`. What are you doing here? You're iterating over the dataframe with some options, providing the `ImageHDU`, applying a user-provided function, then adding some more *new* options. This sounds like you could probably write this function to simply take
```julia
function images(f, df; options_related_to-saving..., kwargs...)
    processed_images = map(images(df; kwargs...)) do img
        # do your thing
        return processed_img
    end
    return processed_images
end
```

Hopefully that makes sense! This will be *invaluable* if you ever add an extra keyword argument to the looping-only functions and don't want to have to rewrite all of these as well! The basic idea is that all these functions do is the same iteration but with some fun file-saving mechanics. Everything else should work the same.

src/collection.jl
```diff
@@ -9,9 +9,66 @@ end
...
+    length(idxs) == 0 && return (filename, "")
+    breaking_index = first(last(idxs))
+    return filename[1:breaking_index - 1], filename[breaking_index:end]
+end
+
+
+"""
+    write_data(file_path, data)
+
+Writes `data` in FITS format at `file_path`.
```
mileslucas, line 48: I would add a comment explaining "why", similar to what should be around `getdata`

src/collection.jl
```diff
@@ -115,30 +172,107 @@ end
...
+        processed_image
+    end
+
+    return processed_images
+end
+
+"""
+    filenames(f, collection::DataFrame; save = false, path = nothing, save_prefix = nothing, save_suffix = nothing, save_delim = "_", ext = r"fits(\\.tar\\.gz)?"i)
+
+Applies function `f` on all file paths present in data frame and saves it in FITS file.
```
mileslucas: very important! It does *not* always save it in a FITS file. You should write this docstring as "this iterates over the filenames of the collection applying function `f` at each step. ... ... ... The following parameters enable programmatically saving the returned value `f` using [`CCDReduction.write_fits`](@ref)"

src/collection.jl
```diff
@@ -88,25 +145,25 @@ function fitscollection(basedir::String;
...
-    return df
+    return collection
 end
 
 
 """
-    arrays(df::DataFrame)
+    arrays(collection::DataFrame)
 
 Generator for arrays of images of entries in data frame.
```
mileslucas, line 159: These docstrings are much too sparse. It should be clear to the user exactly *what* is being output by the iteration, *how* the iteration is occurring, and *what* the options are, if any. Be very clear that this function returns data loaded by FITSIO using `CCDReduction.get_data` into an `Array`. It utilizes the `path` and `hdu` from the collection. 

This will really make clear how these 3 functions are different, as well as how each one operates.

mileslucas (changes requested):

src/collection.jl
```diff
@@ -9,9 +9,72 @@ end
...
+    if !isnothing(save_suffix)
+        modified_name = string(modified_name, save_delim, save_suffix)
+    end
+
+    # adding extension to modified_name
+    if ext == ""
+        file_path = joinpath(save_location, modified_name * ".fits")
+    else
+        file_path = joinpath(save_location, modified_name * ext)
+    end
```
mileslucas: I don't think you should have this branch here. By example, if my filename is `test.fits` and I explicitly call `ext="fizz"` or something, then line 23 will output `modified_name == "test.fits", ext == ""`. then, later at line 31, the output name will be `"test.fits.fits"`. 

Basically you've written code that says "if *you as the user* have given me a bad `ext`, I will add `.fits` to your filename". That doesn't make sense, at least to me! A good idea of how to test this, is that *no matter what `filename` and `ext` are*, if none of the prefix/suffix etc. are provided your output should be exactly the same as the input.

By the way, shadowing variable names, like you do here with `ext` can be *extremely* confusing- I would strongly suggest against it

src/collection.jl
```diff
@@ -9,9 +9,72 @@ end
...
+    return filename[1:breaking_index - 1], filename[breaking_index:end]
+end
+
+
+#=
+FITSIO.jl takes over memory write in by cfitsio, which writes in row-major form,
+whereas when Julia gives that memory, it is assumed as column major.
+Therefore all data written by `write` is transposed.
+Related comment: https://github.com/JuliaAstro/CCDReduction.jl/pull/16#issuecomment-638492572
+=#
```
mileslucas: why not put it in the docstring?

src/collection.jl
```diff
@@ -88,25 +151,44 @@ function fitscollection(basedir::String;
...
+
+Iterative version of `arrays`, returns data loaded by `FITSIO` using `CCDReduction.getdata` into an `Array`.
+It utilizes the `path` and `hdu` from the `collection`.
+
+This version can be used as
+```julia
+collection = fitscollection("~/data/tekdata")
+image_arrays = arrays(collection) |> collect
+```
+this returns all image arrays present in `collection`. This can also be used via a for-loop
```
mileslucas: I would put the examples under an `# Examples` header.

"image array" is kind of weird terminology. You already say what is being returned above, I would just remove it.

src/collection.jl
```diff
@@ -88,25 +151,44 @@ function fitscollection(basedir::String;
...
+```julia
+collection = fitscollection("~/data/tekdata")
+processed_images = Vector{Array}(undef, first(size(collection)))
+for (i, arr) in enumerate(arrays(collection))
+    processed_images[i] = trim(arr, (:, 1040, 1059))
+end
```
mileslucas: perhaps it is just me, but I feel like this example appears complicated without really explaining what the actual function is doing. `Vector{Array}(undef, first(size(collection)))` is pretty intimidating. A much clearer example might look like
```julia
collection = fitscollection("...")
for arr in arrays(collection)
    @assert arr isa Array
    println(size(arr))
end

# output
(1048, 1068)
(1048, 1068)
...
```

Notice how this enforces that `arrays()` returns `Array`s, shows that you *have* to use `fitscollection`, and shows using it in a for-loop without the unnecessary (and scary) `Vector{Array}(undef, first(size(collection)))` and `enumerate` calls.

mileslucas: I would say if you want to stick to the specific example you could just write `trimmed = [trim(arr, (:, 1040:1059)) for arr in arrays(collection)]`.

src/collection.jl
```diff
@@ -88,25 +151,44 @@ function fitscollection(basedir::String;
...
 
 
 """
-    arrays(df::DataFrame)
+    arrays(collection::DataFrame)
 
 Generator for arrays of images of entries in data frame.
+
+Iterative version of `arrays`, returns data loaded by `FITSIO` using `CCDReduction.getdata` into an `Array`.
+It utilizes the `path` and `hdu` from the `collection`.
```
```suggestion
Iterates over `collection` using each `path` and `hdu` to load data using [`CCDReduction.getdata`](@ref) into an `Array`.
```

src/collection.jl
```diff
@@ -88,25 +151,44 @@ function fitscollection(basedir::String;
...
+this returns all image arrays present in `collection`. This can also be used via a for-loop
+```julia
+collection = fitscollection("~/data/tekdata")
+processed_images = Vector{Array}(undef, first(size(collection)))
+for (i, arr) in enumerate(arrays(collection))
+    processed_images[i] = trim(arr, (:, 1040, 1059))
+end
+```
+
+This version involving for-loop (i.e. iterative version) does not support saving functionality.
```
mileslucas: I would remove this. Like I've said before, the saving functionality is *extra*. Write this docstring like it is the de facto method, and in the other docstring add "in addition, by providing blah blah blah you can save the file". This is enough documentation, adding it here basically duplicates the documentation, on top of the fact that the code won't run if you try providing `save_prefix` here, for example.

mileslucas (changes requested):
some docs comments- should propagate to all three versions

src/collection.jl
siddharthlal25: Let's stick with the first example!

src/collection.jl
siddharthlal25: I have used:
```
[`CCDReduction.getdata`](@ref)
[`CCDReduction.write_fits`](@ref)
```
This fails the building of docs, some warning that shows up:
```
┌ Warning: no doc found for reference '[`CCDReduction.write_fits`](@ref)' in src/api.md.
└ @ Documenter.CrossReferences ~/.julia/packages/Documenter/PLD7m/src/CrossReferences.jl:160
┌ Warning: no doc found for reference '[`CCDReduction.write_fits`](@ref)' in src/api.md.
└ @ Documenter.CrossReferences ~/.julia/packages/Documenter/PLD7m/src/CrossReferences.jl:160
┌ Warning: no doc found for reference '[`CCDReduction.write_fits`](@ref)' in src/api.md.
└ @ Documenter.CrossReferences ~/.julia/packages/Documenter/PLD7m/src/CrossReferences.jl:160
```
If I don't cross-reference the documentation, the doc gets build completely. Is it a bug with `Documentation.jl`?

PR #31, opened 2020-07-17
`CCDData` struct to store `ImageHDU`

siddharthlal25:
Last PR #30, in which we tried to make generators for `ImageHDU`s, after the iteration gets over all the FITS handles are closed and the collected `ImageHDU`s are not accessible.

Something like:
```
df = fitscollection("...")
hdus = collect(images(df))
trimmed = [trim(hdu, (:, 1040,1059)) for hdu in hdus]
```
This would error because all handles of read `ImageHDU`s are closed.

So, to solve this we introduce `CCDData` struct. This stores the header and image array, so even if the FITS handles are closed, we will still be able to access the data and header with ease.

Things to tackle in this PR:

- [x] New documentation

- [x] Tests for new Interface

- [x] Support for strings

- [x] Creating a function to create default header

- [x] work on `combine`

Checklist for next PRs:

1. Defining `==` for `CCDData`

2. Iterative and mapping versions of `images`, `filenames` and `arrays` on output of `fitscollection` using `CCDData` interface.

3. Dealing with `Symbols` as keys in header


siddharthlal25:
Should we keep support for `ImageHDU`s or drop it completely? @mileslucas @giordano
At times accessing them can cause errors, because of closed FITS handles.

src/datatype.jl
```diff
@@ -0,0 +1,36 @@
+# custom data tyoe to hold ImageHDU
```
giordano:
```suggestion
# custom data type to hold ImageHDU
```

src/datatype.jl
```diff
@@ -0,0 +1,36 @@
...
+struct CCDData{T,M<:AbstractMatrix{T}} <: AbstractMatrix{T}
+    data::M
+    hdr::FITSHeader
+end
+
+# extending the AbstractMatrix interface
+Base.size(A::CCDData) = size(A.data)
+Base.getindex(A::CCDData, i) = getindex(A.data, i)
+Base.setindex!(A::CCDData, v, i) = setindex!(A.data, v, i)
```
giordano: The `*index*` functions have many methods, not just with a single index

src/datatype.jl
```diff
@@ -0,0 +1,36 @@
...
+Base.eachindex(A::CCDData) = eachindex(A.data)
+Base.iterate(A::CCDData) = iterate(A.data, state=(eachindex(A),))
```
giordano: Shouldn't this method have a `state` optional argument?

giordano:
>  At times accessing them can cause errors, because of closed FITS handles.

Closed or not closed?  Anyway, this sounds more like an error to fix rather than try to avoid?

siddharthlal25:
> > At times accessing them can cause errors, because of closed FITS handles.
> 
> Closed or not closed? Anyway, this sounds more like an error to fix rather than try to avoid?

Closed can cause errors in the example I mentioned above, where one collects ImageHDU using `images`, and then try to apply function on collected ImageHDU (this will cause error since the FITS file has been closed from which the ImageHDU was taken).

src/datatype.jl
siddharthlal25: Are you referring to this one?
```
setindex!(A, v, I::Vararg{Int, N})
```

src/datatype.jl
siddharthlal25: Yes, it should have one. I will add it.

src/datatype.jl
giordano: I'd do
```suggestion
Base.getindex(A::CCDData, inds...) = getindex(A.data, inds...)
Base.setindex!(A::CCDData, v, inds...) = setindex!(A.data, v, inds...)
```

mileslucas:
@giordano The big problem is that after you close the file handle you can't read the data or header from the ImageHDU anymore. This screws up a lot of the future plans I had envisioned for processing, which require a container that holds the data and memory. I had thought ImageHDU would be good enough but it looks like we need an in-memory struct instead. 

However, the code currently works, so there isn't a lot of reason to delete it. I think the choice depends on whether you want to keep writing functions for ImageHDU (in addition to string and array versions). I think the usage is going to be low so it's more of a "do you want to keep developing this"?

siddharthlal25:
I have no problem developing this package, the main problem (that I think), if we keep `ImageHDU` in the pipeline is the trouble with open and closed handles of `FITS` files. Suppose if a user is playing around with some `ImageHDU` and somehow he closes the `FITS` file, this will lead to lot of errors in the subsequent processing code and suppose he leaves the handle open and tries to overwrite then there might be an issue as we saw for Windows. So, why not give the users a single point of action, being `CCDData`, where the user can work with it even with closed `FITS` handles. Thoughts?

mileslucas:
> So, why not give the users a single point of action, being CCDData

Sounds like a great approach!

src/datatype.jl
```diff
@@ -0,0 +1,36 @@
+# custom data tyoe to hold ImageHDU
+struct CCDData{T,M<:AbstractMatrix{T}} <: AbstractMatrix{T}
+    data::M
+    hdr::FITSHeader
+end
```
mileslucas: let's do
```julia
abstract type AbstractCCDData{T} <: AbstractMatrix{T}
struct CCDData{T,M<:AbstractMatrix{T}} <: AbstractCCDData{T}

end
```

This will allow future users to be able to easily extend and make their own `CCDData` type!

src/datatype.jl
mileslucas: you should be able to get away with
```julia
Base.iterate(A::CCDData) = iterate(A.data)
Base.iterate(A::CCDData, state) = iterate(A.data, state)
```

src/datatype.jl
```diff
@@ -0,0 +1,36 @@
...
+Base.size(A::CCDData) = size(A.data)
+Base.getindex(A::CCDData, i) = getindex(A.data, i)
+Base.setindex!(A::CCDData, v, i) = setindex!(A.data, v, i)
+Base.eachindex(A::CCDData) = eachindex(A.data)
+Base.iterate(A::CCDData) = iterate(A.data, state=(eachindex(A),))
+Base.length(A::CCDData) = length(A.data)
+Base.axes(A::CCDData) = axes(A.data)
+Base.IndexStyle(::CCDData) = IndexCartesian()
+
+read_hdu(hdu::ImageHDU) = CCDData(getdata(hdu), read_header(hdu))
```
mileslucas: I think this function name should just be `CCDData` since it's acting as a constructor
```suggestion
CCDData(hdu::ImageHDU) = CCDData(getdata(hdu), read_header(hdu))
```

src/datatype.jl
```diff
@@ -0,0 +1,36 @@
...
+
+
+# defining basic functions on CCDData
+# subtract_bias
+function subtract_bias!(frame::CCDData, bias_frame::CCDData)
+    subtract_bias!(frame.data, bias.data)
+    return frame
+end
```
mileslucas: you shouldn't need this function, right? If `CCDData <: AbstractArray` (and the implementation is correct) you shouldn't have to define any of these functions!

siddharthlal25:
Let's get rid of `ImageHDU`s, should we support `Arrays` with `CCDData` (like bias subtraction on `Array` from `CCDData`) or should we keep things exclusive to `CCDData`?

Edit:
The support would already be available due to `AbstractMatrix` interface!

src/datatype.jl
siddharthlal25: Hmm, interesting! Didn't see that support coming in, that would simplify a lot of things!
I mean we just need to rewrite a bit of generator function and copy the mapping version from the previous PR. That makes the code so much simple! Thank you.

mileslucas:
I had a conversation with Mosè and I think a good idea moving forward for handling headers within our functions looks like this:

Since `CCDData <: AbstractArray` that means we don't *have* to write specialized methods, so how will we be able to do things like `subtract_overscan(data, idxs::Symbol)` to extract the value from the header. This requires us to write many specialized methods. Also, users can only parse the specific things we allow, as opposed to generically supporting substitution.

An idea I had to remedy this is to use a macro for extracting and replacing header data in a function call. Consider this:
```julia
img = CCDData(...)
@hdr img subtract_overscan(img, :DATASEC)
```
where the first `img` is to tell the macro where to get the header, and the rest is some generic expression where symbols will try and be parsed as keys. This would transform the expression to something like this-
```julia
subtract_overscan(img, img.hdr["DATASEC"])
```

This gives the user full control over the substitution, in addition to greatly reducing the amount of code we have to write! This macro is very similar to the `@df` macro used by StatsPlots, so I will take a look there and try and reason with it. For now, if you think this is a good way forward, then you can actually remove those special methods we made (like `subtract_overscan(..., ::Symbol)`) and don't have to add any more new ones.

mileslucas:
Another thing I think worth doing is having a more direct interface to the header. Example usage like
```julia
ccd = CCDData(...)
idxs = ccd[:DATASEC]
ccd["DATASEC"] == idxs
ccd["DATASEC"] = "1:20,1:1040"
```

To do this, I believe all you have to make are
```julia
Base.getindex(ccd::CCDData, key::AbstractString) = ...
Base.getindex(ccd::CCDData, key::Symbol) = ...
Base.getindex(ccd::CCDData, keys::AbstractVector{<:AbstractString}) = ...
Base.getindex(ccd::CCDData, keys::AbstractVector{Symbol}) = ...
```
and the equivalent `setindex!` functions!

You may also want to add `Base.keys` `Base.values`, and some other methods from the `AbstractDict` interface.

siddharthlal25:
The idea of using macros for header parsing sounds very interesting! I will definitely go with it. I will also look into `@df` from StatsPlots.

siddharthlal25:
Some trouble:
```
frame = CCDData(fill(2, 5, 5), hdr)
bias = CCDData(fill(2.0, 5, 5), hdr)
proc = subtract_bias(frame, bias) # works completely fine

frame = CCDData(fill(2, 5, 5), hdr)
bias = CCDData(fill(2.5, 5, 5), hdr)
proc = subtract_bias(frame, bias) # Inexact error
```
We need to tackle the case when types are different (i.e. of `frame` and `bias`) for most of the functions, so I tried the way we resolved this for `flat_correct`: (i.e. type promotion)
```
function flat_correct(frame::AbstractArray{T}, flat_frame::AbstractArray{S}; kwargs...) where {T, S}
    V = float(promote_type(T, S))
    return flat_correct!(V.(frame), flat_frame; kwargs...)
end
```
But since we have `V.(frame)` (this becomes an `Array`), because it copies only the `Array` part in `frame` before calling `flat_correct!`, so the final returned value will be an `Array` even though we had passed `CCDData`, I have thought of the following way to tackle this:

Take the part of type promotion to mutating versions and then define a custom version of `copy`, so that doing `V.(frame)` still results in `CCDData` and then function routine remains the same thereafter. (this will support `flat_correct!` with `frame` and `flat_frame` having different types and this can be extended to `subtract_bias` as well) 

What would be your take on this? Are there any better ways to tackle this issue? @giordano @mileslucas

mileslucas:
A couple things:

1. Yes we need to figure out type promotion, but we can do it at the interface level
Let's start by defining a type-ful constructor
```julia
CCDData{T}(data, hdr) where {T<:Number} = CCDData(T.(data), hdr)
```
we're also going to want a constructor that doesn't require `hdr`, so you'll want to write a function that creates a default header given an image. This is described [in this astropy code](https://github.com/astropy/astropy/blob/d1e34fb0c9cff97dc1b7efd024cea2485440c832/astropy/io/fits/hdu/image.py#L35) I would just make something like this
```julia
CCDData(data) = CCDData(data, default_fits_header(data))
```

from here, we should be able to just define some promotion rules using `promote_rule` ([docs](https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#Defining-Promotion-Rules-1))
```julia
Base.promote_rule(::Type{CCDData{T}}, ::Type{CCDData{V}}) where {T,V} =
    CCDData{promote_type{T,V}}
```
start here, and see if that is enough to fix the problem.

2. We need to figure out how to avoid `T.(frame)` from turning `frame` into an `Array`

Turns out this is almost a direct copy-paste from https://docs.julialang.org/en/v1/manual/interfaces/#Selecting-an-appropriate-output-array-1 I've pushed directly to add this
```julia
# broadcast mechanics
Base.BroadcastStyle(::Type{<:CCDData}) = Broadcast.ArrayStyle{CCDData}()
function Base.similar(bc::Broadcast.Broadcasted{Broadcast.ArrayStyle{CCDData}}, ::Type{T}) where T
    ccd = find_ccd(bc)
    CCDData(similar(ccd.data, T, axes(bc)), ccd.hdr)
end
"`A = find_ccd(As)` returns the first CCDData among the arguments."
find_ccd(bc::Base.Broadcast.Broadcasted) = find_ccd(bc.args)
find_ccd(args::Tuple) = find_ccd(find_ccd(args[1]), Base.tail(args))
find_ccd(x) = x
find_ccd(::Tuple{}) = nothing
find_ccd(a::CCDData, rest) = a
find_ccd(::Any, rest) = find_ccd(rest)
```

giordano:
Now that we have `AbstractCCDData`, I'd define all methods for this type.  This would make it easy for other users to define their own CCD data type which has only some differences with the reference implementation, and also helps making sure we define a consistent interface.

mileslucas:
how would you write `Base.size(::AbstractCCDData)` for example?

giordano:
> how would you write Base.size(::AbstractCCDData) for example?

Ok, maybe not _all_, but most of the methods that can be common to all "CCD data types" (and users can always define a more specific method for a custom subtype)

mileslucas:
Right, for everything else that doesn't require underlying knowledge we can define for the abstract type. We can even define size for the abstract type and it will just fail fast for any sub types so they would know to write their own.

giordano:
> Right, for everything else that doesn't require underlying knowledge we can define for the abstract type.

Exactly.  `size` depends on the specific implementation, so it's correct that we define it for the concrete type

siddharthlal25:
I think this will still be an issue, suppose we do this:
```
frame = CCDData(fill(2, 5, 5), hdr)
bias = CCDData(fill(2.5, 5, 5), hdr)
subtract_bias!(frame, bias) # Inexact error
```
Now this will throw an inexact error since there is type mutation after bias subtraction, should we tackle this? Or should it be taken care of from the user's end?

The promotion rule will work pretty well with the non-mutating version!

giordano:
> Or should it be taken care of from the user's end?

Looks like a user's problem to me

siddharthlal25:
> Looks like a user's problem to me

Cool! I will add error tests for these as well! Thank you so much!

mileslucas:
Another piece of code I believe you'll need for getting promotion to work is
```julia
Base.convert(::Type{CCDData{T}}, ccd::CCDData{V}) where {T,V} = CCDData{T}(ccd.data, ccd.hdr) 
```
and potentially a few more, but at minimum this convert method

siddharthlal25:
What should we do about `combine`, the current implementation uses `LazyStack`:
```
function combine(frames::Vararg{<:AbstractArray{<:Number}}; method = median)
    firstframe = first(frames)
    dim = ndims(firstframe) + 1
    shape = size(firstframe)
    return reshape(method(LazyStack.stack(frames), dims = dim), shape)
end

combine(frames; method = median, kwargs...) = combine(frames...; method = method, kwargs...)
```
And this returns an `Array` even if we pass in `CCDData`, because of `stack(frames)` results in `Array`.

Edit:
for `trim` we rely upon `trimview` which returns `selectdim(frame, d, complement_idxs)`, how do we make sure that the output for `selectdim` is `CCDData` given `frame` is `CCDData`, should we create a new function something like:

```
function selectdim(frame::CCDData, d, idxs)
    frame.data = selectdim(frame.data, d, idxs)
    return frame
end
```
Or can we add something in the interface part and make it work directly? Thoughts?

siddharthlal25:
Some fundamental questions about functionalities, `trimview` returns `view` of an `Array`. With new structs in play, what should we return, should we return a new `CCDData` struct with a view of the passed `CCDData` or only an `Array` that is a `view` of `CCDData.data`?

mileslucas:
Hmm, can you add a method `Base.view` that propagates to the interior, like
```julia
Base.view(ccd::CCDData, inds...) = CCDData(view(ccd.data, inds...), ccd.hdr)
```
?

I'm not sure if this is common practice or even a good idea.

Edit: If you add this, you'd want to symmetrically write a `Base.parent` method in the same way.

siddharthlal25:
Currently, the docs build is breaking because of some documentation in `docs/src/index.md`, will fix it once the code gets ready.

Edit:
Will fix it once I have written the default header generating function.

Edit:
There is a default header generator in `FITSIO.jl`, let me check if we can use it.

Edit:
Yup, there is something which we can use:
```
x = zeros(5, 5)

FITS("master_bias.fits", "w") do f
    write(f, x)
end

dir = joinpath("~/.julia/dev/CCDReduction.jl", "master_bias.fits")

imagehdu = CCDData(FITS(dir)[1])

imagehdu.hdr
# output

SIMPLE  =                    T / file does conform to FITS standard
BITPIX  =                  -64 / number of bits per data pixel
NAXIS   =                    2 / number of data axes
NAXIS1  =                    5 / length of data axis 1
NAXIS2  =                    5 / length of data axis 2
EXTEND  =                    T / FITS dataset may contain extensions
COMMENT   FITS (Flexible Image Transport System) format is defined in 'Astronom
COMMENT   and Astrophysics', volume 376, page 359; bibcode: 2001A&A...376..359H
```

Edit:

I am not able to generate a default header and hence stuck with `combine`. I have thought to keep the generated default header (as in the above example) with the output of `combine` (not inherit the header of the first argument). Thoughts?

src/fits.jl
```diff
@@ -11,3 +11,32 @@ function getdata(hdu::ImageHDU)
...
+"""
+function subtract_bias end
+
+
+"""
+    subtract_bias!(frame, bias_frame)
+
+Load HDU for the `frame` or `bias_frame` as `CCDData` before subtracting from `frame` in-place.
+"""
+function subtract_bias! end
```
siddharthlal25: Thoughts about docs? @mileslucas 
Once this looks good, I will follow the same style to write docs for other functions.

src/CCDReduction.jl
```diff
@@ -22,10 +22,12 @@ export subtract_bias,
...
        arrays,
        filenames,
-       images
+       images,
+       CCDData
 
 include("methods.jl")
 include("fits.jl")
 include("collection.jl")
+include("datatype.jl")
```
mileslucas: I'm not sure this filename describes the contents well. I think `ccddata.jl` is a good name.

src/datatype.jl
```diff
@@ -0,0 +1,35 @@
...
+
+# extending the AbstractMatrix interface
+Base.size(ccd::CCDData) = size(ccd.data)
+Base.getindex(ccd::CCDData, inds...) = getindex(ccd.data, inds...) # default fallback for operations on Array
+Base.setindex!(ccd::CCDData, v, inds...) = setindex!(ccd.data, v, inds...) # default fallback for operations on Array
+Base.promote_rule(::Type{CCDData{T}}, ::Type{CCDData{V}}) where {T,V} = CCDData{promote_type{T,V}}
+Base.convert(::Type{CCDData{T}}, ccd::CCDData{V}) where {T,V} = CCDData{T}(ccd.data, ccd.hdr)
+Base.view(ccd::CCDData, inds...) = CCDData(view(ccd.data, inds...), ccd.hdr)
+Base.selectdim(ccd::CCDData, d::Integer, idxs) = CCDData(selectdim(ccd.data, d, idxs), ccd.hdr)
+Base.copy(ccd::CCDData) = CCDData(copy(ccd.data), deepcopy(ccd.hdr))
```
mileslucas: I feel like both should use `copy` if you're calling `copy`
```suggestion
Base.copy(ccd::CCDData) = CCDData(copy(ccd.data), copy(ccd.hdr))
```

mileslucas (commented):
Looks like a good start- think of what can be written in terms of `AbstractCCDData`. Consider adding formal API for `AbstractCCDData` to get the data and header items, which would allow you to fully implement everything here using generic code, allowing much easier extension of `AbstractCCDData` in the future.

src/fits.jl
mileslucas: The line "The parameters can also be of mixed type, e.g. `frame` is an `Array` and `bias_frame` is `CCDData` or vice versa." is unnecessary I think.

Also consider condensing these separate docs into main methods docstrings, just adding something like "if either are strings they will be loaded into a [`CCDData`](@ref) first. The HDU loaded can be specified by..."

src/datatype.jl
siddharthlal25: `Base.copy` is not defined for `FITSHeader`.

mileslucas: okay, I would open an issue/PR at FITSIO and use `deepcopy` for now.

siddharthlal25:
I have put up a small piece that enables us to get the default header value of an array, next we need to tackle one design problem:
Problem:
What should be the output of `combine`?

The case when all elements are `Arrays` (the trivial one), this returns an `Array`.

Next, the design question:
What should be the output if even one of the input to combine is `CCDData`, should we return a `CCDData`? Or should we return a `CCDData` only when all inputs to `combine` is `CCDData`. Thoughts? @mileslucas @giordano 

Edit:
The doc builds successfully after implementing the `String` supporting version of functions :tada:

siddharthlal25:
Travis is very slow, should we switch to GitHub actions?

mileslucas:
> What should be the output

To begin, I want to point out this functionality is defined by the call to `Base.similar` in the broadcast section. Here is my thinking. For simple operations I think this eagerness is beneficial-
```julia
ccd = CCDData(...)
calib_ccd = (ccd ./ 10) .- minimum(ccd)
calib_ccd isa CCDData # true
hdr(calib_ccd) == hdr(ccd) # true
```
this is pretty natural workflow and the automatic header copying is nice.

For operations like `subtract_overscan` I think it is pretty clear that there is some image of interest that we are manipulating, so we should copy that image's header through, like this:
```julia
ccd = CCDData(...)
calib_ccd = correct_flat(subtract_overscan(ccd, "bias.fits"), "flat.fits")
calib_ccd isa CCDData # true
hdr(calib_ccd) == hdr(ccd) # true
```

Now for the complicated version, which is operations containing multiple images with no clear precedence. First, using basic operators
```julia
ccd = CCDData(...)
bias = CCDData("bias.fits")
flat = CCDData("flat.fits")
calib_ccd = (ccd .- bias) ./ flat
```
the current functionality will basically keep the header from the _first_ CCDData. In the case above, this works perfectly as expected, but it imposes a syntax restriction on the user.

For the `combine` method I think a good default is to copy the first `CCDData` available, with an optional keyword argument to specify the hdr to copy.

siddharthlal25:
The current implementation of `combine` in master leads to a small error:
```
x1 = zeros(4, 5)
x2 = ones(4, 5)

combine(x1, x2) # works fine

x2 = fill(Float32(1), 4, 5)
combine(x1, x2) # errors BoundsError: attempt to access 2-element Array{Any,1} at index [3]
```
I think this happens because of the different types of elements in `x1` and `x2`. So, should we worry about promoting types on our end or leave it for the user? @mileslucas @giordano 

Edit:
Fixed it in [cb71530](https://github.com/JuliaAstro/CCDReduction.jl/pull/31/commits/cb715300c25feb104709d83a268875abf429e91c)

src/ccddata.jl
```diff
@@ -0,0 +1,42 @@
+# abstract data type for CCDData
```
mileslucas, line 1: docs :)

src/ccddata.jl
```diff
@@ -0,0 +1,42 @@
...
+
+CCDData(hdu::ImageHDU) = CCDData(getdata(hdu), read_header(hdu))
+CCDData(data) = CCDData(data, get_default_header(data))
+function CCDData(path::String, hdu)
+    fh = FITS(path)
+    ccd = CCDData(fh[hdu])
+    close(fh)
+    return ccd
+end
+CCDData{T}(data, hdr) where {T<:Number} = CCDData(T.(data), hdr)
```
mileslucas: docs, too.

Project.toml
```diff
@@ -4,13 +4,15 @@ authors = ["Siddharth Lal"]
 version = "0.2.0"
 
 [deps]
+CFITSIO = "3b1b4be9-1499-4b22-8d78-7db3344d1961"
```
mileslucas: ehhh I think we should avoid depending on CFITSIO. Some long-term over-arching goals of FITSIO are to remove the dependency because CFITSIO is a c-library rather than a native julia implementation.

You can get `bitpix_from_type` using FITSIO, you'll just have to upper-bound the compat. Also it's small enough of a method you could just copy and paste it.

src/fits.jl
```diff
@@ -12,173 +12,65 @@ function getdata(hdu::ImageHDU)
...
+
+    # assigning comments
+    comments = ["file does conform to FITS standard",                                   # comment for SIMPLE
+                "number of bits per data pixel",                                        # comment for BITPIX
+                "number of data axes",                                                  # comment for NAXIS
+                [string(Symbol("length of data axis ", i)) for i in 1:ndims(data)]...,  # comments for axis length
+                "FITS dataset may contain extensions"]                                  # comment for EXTEND
+
+    return FITSHeader(hdu_keys, hdu_values, comments)
+end
```
mileslucas: probably more appropriate to PR into FITSIO.jl

src/methods.jl
```diff
@@ -33,10 +35,14 @@ end
 
 
 """
-    subtract_bias(frame::AbstractArray, bias_frame::AbstractArray)
+    subtract_bias(frame, bias_frame; [hdu = (1, 1)])
```
mileslucas: So, I would normally choose to keep as brief a docstring as possible, e.g. `[hdu = 1]`, since that encapsulates the behavior perfectly, and in cases like `combine` it avoids you implying any dimensionality.

This is totally up to your discretion, though, consistency is the most important!

src/methods.jl
```diff
@@ -33,10 +35,14 @@ end
...
 
 Subtract the `bias_frame` from `frame`.
 
+If either are strings, they will be loaded into `CCDData` first. The HDU loaded can be specified by `hdu` as either an integer or a tuple corresponding to each file.
```
```julia
[`CCDData`](@ref)
```

src/methods.jl
```diff
@@ -33,10 +35,14 @@ end
...
+
+Function output inherits the type of first parameter and header file of output (if applicable) is same as of the first parameter.
```
mileslucas: This reads a little confusing- I'm not sure it's really necessary for methods like `subtract_bias` where there is a clear "frame of interest" and a "calibration frame". Ideally, through the function name and the argument names it should be clear what is happening (I think so, although I may be wrong), and everything should just work.

src/methods.jl
```diff
@@ -56,10 +62,12 @@ subtract_bias(frame::AbstractArray, bias_frame::AbstractArray) = subtract_bias!(
 
 
 """
-    subtract_overscan!(frame::AbstractArray, idxs; dims = axes_min_length(idxs))
+    subtract_overscan!(frame, idxs; dims = axes_min_length(idxs))
 
 In-place version of [`subtract_overscan`](@ref)
 
+`frame` cannot be a string for mutating version.
```
mileslucas: Similar to above, to ME it makes sense that you can't mutate on a string, so it shouldn't need an explanation. However, this requires knowledge of the idiosyncrasies of `!` methods, so from that viewpoint its not a terrible idea to explain it. 

If you do explain it, you'll need to be more explicit. The first thing I would saw if I read this is "why?". Unfortunately, this means you'll be copying the same explanation to all the docstrings for consistency.

So I think either remove this or put an extra line explaining why- up to you!

src/methods.jl
```diff
@@ -117,12 +131,16 @@ end
...
 
 """
-    flat_correct(frame::AbstractArray, flat_frame::AbstractArray; norm_value = mean(flat_frame))
+    flat_correct(frame, flat_frame; norm_value = mean(flat_frame), [hdu = (1, 1)])
 
 Correct `frame` for non-uniformity using the calibrated `flat_frame`.
 
 By default, the `flat_frame` is normalized by its mean, but this can be changed by providing a custom `norm_value`.
 
+If either are strings, they will be loaded into `CCDData` first. The HDU loaded can be specified by `hdu` as either an integer or a tuple corresponding to each file.
```
```julia
[`CCDData`](@ref)
```

src/methods.jl
```diff
@@ -194,12 +216,14 @@ trim(frame, idxs) = copy(trimview(frame, idxs))
...
 
 """
-    trimview(frame::AbstractArray, idxs)
+    trimview(frame, idxs)
 
 Trims the `frame` to remove the region specified by idxs.
 
 This function is same as the [`trim`](@ref) function but returns a view of the frame.
 
+`frame` cannot be a string. Function output inherits the type of first parameter and header file of output (if applicable) is same as of the first parameter.
```
mileslucas: combo of both my previous comments! The first half makes me ask "why" and the second half is a little hard to read/hard to understand.

src/methods.jl
```diff
@@ -305,13 +335,17 @@ end
...
-    combine(frames...; method = median)
-    combine(frames; method = median)
+    combine(frames...; method = median, [hdu = 1], [header_hdu = 1])
+    combine(frames; method = median, [hdu = 1], [header_hdu = 1])
 
 Combine multiple frames using `method`. Multiple frames can also be passed in a vector or as generators for combining.
 
 To pass a custom method, it must have a signature like `method(::AbstractArray; dims)`.
 
+If `frames` is strings, , it will be loaded into `CCDData` first. The HDU number can be specified with `hdu` as either an integer or a tuple corresponding to each file.
```
```suggestion
If `frames` are strings, they will be loaded into [`CCDData`](@ref)s first. The HDU indices can be specified with `hdu` as either an integer or a tuple corresponding to each file.
```

src/methods.jl
```diff
@@ -332,17 +366,19 @@ function combine(frames::Vararg{<:AbstractArray{<:Number}}; method = median)
     firstframe = first(frames)
     dim = ndims(firstframe) + 1
     shape = size(firstframe)
-    return reshape(method(LazyStack.stack(frames), dims = dim), shape)
+    return reshape(method(LazyStack.stack(frames...), dims = dim), shape)
 end
 
-combine(frames; method = median, kwargs...) = combine(frames...; method = method, kwargs...)
+combine(frames; kwargs...) = combine(frames...; kwargs...)
```
mileslucas, line 352: I think this creates ambiguities if I were to call `combine(randn(512, 512))` ie combine one frame. I think this behavior should be allowed- if it doesn't work as is you'll just have to play around with typing the methods.

mileslucas (changes requested):
Most of the code looks decent so far, mostly lacking CCDData docs and we need to find where to put the header parsing stuff. Good work so far!

Project.toml
siddharthlal25: Let's upper bound the `compat`.

src/methods.jl
siddharthlal25: This one is not ambiguous:
```
julia> @which combine(randn(512, 512))
combine(frames::AbstractArray{#s27,N} where N where #s27<:Number...; method) in CCDReduction at /home/siddharth/.julia/dev/CCDReduction.jl/src/methods.jl:366
```
Goes to the `Vararg` version (because of `AbstractArray{<:Number}` typing)

src/fits.jl
siddharthlal25: Let me look into it.

src/methods.jl
siddharthlal25: Since this is not the in-place version, things can be a bit ambiguous about the header.

User thoughts:
1. I am allocating a new `CCDData`, it will use the constructor `get_default_header` to get the header of the newly created struct.
2. I am allocating a new `CCDData`, it will have the same header as the frame of interest.

Therefore, I explicitly mentioned the source of the new header created. Thoughts?

src/methods.jl
siddharthlal25: Let's remove this, explaining why strings are not supported would make the docs quite verbose, which may intimidate the user :laughing:

src/methods.jl
siddharthlal25: Let's use `[hdu = 1]`, this looks more appropriate!

mileslucas:
I don’t understand what you’re saying about the header creation and
CCDData.

src/methods.jl
siddharthlal25: When a user goes for a non-mutating function, the method creates a new instance of struct `CCDData`, right?

Now, how will the user know, whether the header which was in output (new struct) was copied from "calibration frame" or generated using the `get_default_header`, are you following?

mileslucas:
Neither. If “frame” is an array and “bias” is a CCDData the output should
be an array.

src/methods.jl
siddharthlal25: To tackle `(::Array, ::CCDData)` (output will be an `Array`) case, I have mentioned the applicability:
"header file of output (if applicable) is same as of the first parameter."

docs/Project.toml
```diff
@@ -5,5 +5,5 @@ Underscores = "d9a01c3f-67ce-4d8c-9b55-35f6e4050bb1"
 
 [compat]
 Documenter = "0.24"
-FITSIO = "0.14"
+FITSIO = "0.15.0"
```
mileslucas: I would suggest this to allow patch compatibility
```suggestion
FITSIO = "0.15"
```

src/CCDReduction.jl
```diff
@@ -22,8 +22,10 @@ export subtract_bias,
        fitscollection,
        arrays,
        filenames,
-       images
+       images,
```
mileslucas, line 25: maybe consider changing `images` to something else? Originally it was `images` because it generated `ImageHDU`s. Now it is generating `CCDData`, perhaps we rename to `ccds` (this is what ccdproc does), `data` (pretty ambiguous), `ccddata`?

src/ccddata.jl
```diff
@@ -0,0 +1,69 @@
...
+
+Supertype for `CCDData` based on `AbstractMatrix` interface.
+"""
+abstract type AbstractCCDData{T} <: AbstractMatrix{T} end
+
+# custom data type to hold ImageHDU
+"""
+    CCDData(data::AbstractMatrix, hdr::FITSHeader)
+
+Struct to store `ImageHDU`, derived from [`AbstractCCDData`](@ref).
```
mileslucas, line 25: This is definitely not enough documentation for this struct. Some examples showing how it works (show how it is basically just an array!!), explain how headers are copied over/generated (very important! Show an example like `a + b` vs `b + a` returning different headers). Also, you don't need a docstring for every constructor, for example the constructor `CCDData(::AbstractMatrix)` can be shown in this docstring by just putting
```julia
"""
    CCDData(data, [hdr])
```
and explaining "if `hdr` is not provided, a default will be constructed based on the data."

src/ccddata.jl
```diff
@@ -0,0 +1,69 @@
...
+
+"""
+    CCDData(data::AbstractArray)
+
+Creates `CCDData` from an `Array`.
+"""
+CCDData(data) = CCDData(data, get_default_header(data))
+
+"""
+    CCDData(path::String, hdu)
```
mileslucas: `hdu` should be default to `1`. This is the norm.

src/ccddata.jl
```diff
@@ -0,0 +1,69 @@
...
+
+Loads HDU from `hdu` index in `FITS` file at `path` as `CCDData`.
+"""
+function CCDData(path::String, hdu)
+    fh = FITS(path)
+    ccd = CCDData(fh[hdu])
+    close(fh)
```
mileslucas: style, but this entire thing can be one-lined.
```julia
CCDData(path::AbstractString, hdu) = FITS(f -> CCDData(f[hdu]), path)
```
also note `AbstractString` vs `String`

src/ccddata.jl
```diff
@@ -0,0 +1,69 @@
...
+    return ccd
+end
+CCDData{T}(data, hdr) where {T<:Number} = CCDData(T.(data), hdr)
+
+# extending the AbstractMatrix interface
+Base.size(ccd::CCDData) = size(ccd.data)
```
mileslucas: from this line on down, you should probably switch to using `AbstractCCDData`. As I wrote before, if you define the interface like this
```julia
data(ccd::CCDData) = ccd.data
hdr(ccd::CCDData) = ccd.hdr
```
you should have no problems.

E.G

```suggestion
Base.size(ccd::AbstractCCDData) = size(data(ccd))
```


If you don't do this, then `AbstractCCDData` is essentially worthless, as no methods have been defined for it, meaning any downstream users will have to completely implement _the exact same code that is being implemented here_.

src/methods.jl
mileslucas: that language is still really hard to parse, in my opinion! A different way of saying the same thing is "if `frame` is an `AbstractCCDData` the output will copy `frame`'s header". To me, this seems like the expected behavior, so it's weird to explicitly write out the documentation, but I guess it can't hurt.

src/CCDReduction.jl
siddharthlal25: Let's go with `ccds`, that would make life easier for the users that transition from ccdproc! Also, let's do this in the next PR, as in we can play along with generator code and then the saving versions in a clean piece of code.
PS: The diff seems quite messy.

src/ccddata.jl
siddharthlal25: Sure, will get that done!

src/methods.jl
siddharthlal25: Ya, that looks much better than the current one!

src/ccddata.jl
siddharthlal25: Sure, on it!

src/fits.jl
siddharthlal25: Opened a PR https://github.com/JuliaAstro/FITSIO.jl/pull/140

src/methods.jl
siddharthlal25: Okay, let's remove it from here and show some functionalities of `a + b` in `CCDData` struct docs, we can specify the header stuff over there.

src/ccddata.jl
siddharthlal25: I think the functions `copy`, `convert`, `view`, `selectdim` and broadcast mechanics cannot be generalised for `AbstractCCDData` (let me know if that is possible). I would like to know your thoughts about the current implementation, what more modifications can we do?
I tried to generalise broadcast mechanics and that leads to a lot of errors.

Edit:
pushed the changes in [38d1213](https://github.com/JuliaAstro/CCDReduction.jl/pull/31/commits/38d1213ece08f4591c9778d4a48705b55a0307d8)

src/ccddata.jl
```diff
@@ -14,9 +14,34 @@ Base.promote_rule(::Type{AbstractCCDData{T}}, ::Type{AbstractCCDData{V}}) where
...
+ccd2 = CCDData(ones(4, 4))
+
+sum_ccd1 = ccd1 + ccd2
+
+sum_ccd2 = ccd2 + ccd1
+```
+
+`sum_ccd1` has the header of `ccd1` whereas `sum_ccd2` has the header `ccd2`.
+
+If header is not provided in the `CCDData` constructor, `default_header` is used to generate the header.
```
siddharthlal25, line 60: How can we add `(@ref)` to link documentation of other packages?

src/ccddata.jl
siddharthlal25: Pushed some docs in [d60e235](https://github.com/JuliaAstro/CCDReduction.jl/pull/31/commits/d60e235940b7c1250dec90ddd6b22c621ce671ef), let me know your thoughts about it. Thank you!

src/ccddata.jl
mileslucas: you can't. Typically you can link to the package url or just don't link it.

src/ccddata.jl
mileslucas: Yep, those things are probably going to unavoidable. What are the errors?

siddharthlal25: All lines involving concrete types errors:
```
Base.copy(ccd::CCDData) = CCDData(copy(data(ccd)), deepcopy(hdr(ccd)))
Base.convert(::Type{CCDData{T}}, ccd::CCDData{V}) where {T,V} = CCDData{T}(data(ccd), hdr(ccd))
Base.view(ccd::CCDData, inds...) = CCDData(view(data(ccd), inds...), hdr(ccd))
Base.selectdim(ccd::CCDData, d::Integer, idxs) = CCDData(selectdim(data(ccd), d, idxs), hdr(ccd))
```
and also the broadcast mechanics part:
```
function Base.similar(bc::Broadcast.Broadcasted{Broadcast.ArrayStyle{CCDData}}, ::Type{T}) where T
    ccd = find_ccd(bc)
    CCDData(similar(data(ccd), T, axes(bc)), hdr(ccd))
end
```
Because we are creating a new `CCDData` inside it, so we cannot substitute `AbstractCCDData` in place of `CCDData`

The errors were mostly of the form:
```
No methods found
AbstractCCDData(view(...), ::FITSHeader)
```
Same for `selectdim`, `copy` and `convert`

src/ccddata.jl
siddharthlal25: Do we link it like [abc](www.google.com)?

I tried this, but the REPL output for help gets messed up. (the link also becomes visible)

mileslucas: by default the docstrings are going to show the full URL, not just a blue-text like you see with `@ref`, so that's the expected output.

src/ccddata.jl
mileslucas: okay, that sounds about right, but give this a try for the broadcasting:
```julia
function Base.similar(bc::Broadcast.Broadcasted{Broadcast.ArrayStyle{CCD}}, ::Type{T}) where {T,CCD<:AbstractCCDData}
    ccd = find_ccd(bc)
    CCD(similar(data(ccd), T, axes(bc)), hdr(ccd))
end
```

siddharthlal25: This part works pretty well:
```
Base.BroadcastStyle(::Type{<:CCDData}) = Broadcast.ArrayStyle{CCDData}()
function Base.similar(bc::Broadcast.Broadcasted{Broadcast.ArrayStyle{CCD}}, ::Type{T}) where {T, CCD<:AbstractCCDData}
    ccd = find_ccd(bc)
    CCD(similar(data(ccd), T, axes(bc)), hdr(ccd))
end
"`A = find_ccd(As)` returns the first CCDData among the arguments."
find_ccd(bc::Base.Broadcast.Broadcasted) = find_ccd(bc.args)
find_ccd(args::Tuple) = find_ccd(find_ccd(args[1]), Base.tail(args))
find_ccd(x) = x
find_ccd(::Tuple{}) = nothing
find_ccd(a::AbstractCCDData, rest) = a
find_ccd(::Any, rest) = find_ccd(rest)
```

I had trouble generalising this line:
```
Base.BroadcastStyle(::Type{<:CCDData}) = Broadcast.ArrayStyle{CCDData}()
```

I tried this:
```
Base.BroadcastStyle(::Type{T}) where {T<:AbstractCCDData} = Broadcast.ArrayStyle{T}()
```
But that did not work. Any work around for this?

mileslucas (approved):

src/ccddata.jl
siddharthlal25: For now, let's keep broadcast mechanics specific to subtype, we can discuss about this later!

siddharthlal25:
Thank you so much @mileslucas and @giordano for bearing with me in this PR. 😄 
Finally merging this one! 🎉

PR #33, opened 2020-08-06
Save compatible version of `filenames`, `arrays` and `ccds` function

siddharthlal25:
This PR will mostly be taken from #30 with some minor modifications.

### DIFF
--- a/src/collection.jl
+++ b/src/collection.jl
@@ -42,8 +42,8 @@ end
 
 
 """
-    writefits(file_path, data; header = nothing)
-    writefits(file_path, ccd::CCDData)
+    CCDReduction.writefits(file_path, data; header = nothing)
+    CCDReduction.writefits(file_path, ccd::CCDData)
 
 Writes `data`/`ccd` in FITS format at `file_path`.
 
@@ -250,7 +250,15 @@ end
 
 
 """
-    ccds(f, collection; save = any(!isnothing, (save_prefix, path, save_suffix)), path = nothing, save_prefix = nothing, save_suffix = nothing, save_delim = "_", ext = r"fits(\\.tar\\.gz)?"i, kwargs...)
+    ccds(f,
+         collection;
+         path = nothing,
+         save_prefix = nothing,
+         save_suffix = nothing,
+         save = any(!isnothing, (save_prefix, path, save_suffix)),
+         save_delim = "_",
+         ext = r"fits(\\.tar\\.gz)?"i,
+         kwargs...)
 
 Iterates over the `CCDData`s of the collection applying function `f` at each step.
 
@@ -270,7 +278,7 @@ The above generates `processed_images` which consists of trimmed versions of ima
 
 For saving the `processed_images` simultaneously with the operations performed
 ```julia
-processed_images = map(ccds(collection; save = true, path = "~/data/tekdata", save_prefix = "trimmed")) do img
+processed_images = map(ccds(collection; path = "~/data/tekdata", save_prefix = "trimmed")) do img
     trim(img, (:, 1040:1059))
 end
 ```
@@ -278,7 +286,15 @@ The trimmed images are saved as `trimmed_(original_name)` (FITS files) at `path 
 
 Mapping version of `ccds` function is interfaced on iterative version of `images`, any valid parameter can be passed into iterative version as `kwargs`.
 """
-function ccds(f, collection; save = any(!isnothing, (save_prefix, path, save_suffix)), path = nothing, save_prefix = nothing, save_suffix = nothing, save_delim = "_", ext = r"fits(\.tar\.gz)?"i, kwargs...)
+function ccds(f,
+              collection;
+              path = nothing,
+              save_prefix = nothing,
+              save_suffix = nothing,
+              save = any(!isnothing, (save_prefix, path, save_suffix)),
+              save_delim = "_",
+              ext = r"fits(\.tar\.gz)?"i,
+              kwargs...)
     image_iterator = ccds(collection; kwargs...)
     locations = collection.path
 
@@ -296,7 +312,15 @@ end
 
 
 """
-    filenames(f, collection; save = any(!isnothing, (save_prefix, path, save_suffix)), path = nothing, save_prefix = nothing, save_suffix = nothing, save_delim = "_", ext = r"fits(\\.tar\\.gz)?"i, kwargs...)
+    filenames(f,
+              collection;
+              path = nothing,
+              save_prefix = nothing,
+              save_suffix = nothing,
+              save = any(!isnothing, (save_prefix, path, save_suffix)),
+              save_delim = "_",
+              ext = r"fits(\\.tar\\.gz)?"i,
+              kwargs...)
 
 Iterates over the file paths of the collection applying function `f` at each step.
 
@@ -318,7 +342,7 @@ end
 The above generates `data` which consists of image arrays corresponding to 1st hdu of FITS file paths present in `collection`.
 For saving the `data` simultaneously with the operations performed
 ```julia
-data = map(filenames(collection; save = true, path = "~/data/tekdata", save_prefix = "retrieved_from_filename")) do img
+data = map(filenames(collection; path = "~/data/tekdata", save_prefix = "retrieved_from_filename")) do img
     fh = FITS(path)
     data = getdata(fh[1]) # assuming all 1-hdu are ImageHDUs
     close(fh)
@@ -329,7 +353,15 @@ The retrieved data is saved as `retrieved_from_filename_(original_name)` (FITS f
 
 Mapping version of `filenames` function is interfaced on iterative version of `filenames`, any valid parameter can be passed into iterative version as `kwargs`.
 """
-function filenames(f, collection; save = any(!isnothing, (save_prefix, path, save_suffix)), path = nothing, save_prefix = nothing, save_suffix = nothing, save_delim = "_", ext = r"fits(\.tar\.gz)?"i, kwargs...)
+function filenames(f,
+                   collection;
+                   path = nothing,
+                   save_prefix = nothing,
+                   save_suffix = nothing,
+                   save = any(!isnothing, (save_prefix, path, save_suffix)),
+                   save_delim = "_",
+                   ext = r"fits(\.tar\.gz)?"i,
+                   kwargs...)
     path_iterator = filenames(collection; kwargs...)
     locations = collection.path
 
@@ -347,7 +379,15 @@ end
 
 
 """
-    arrays(f, collection; save = any(!isnothing, (save_prefix, path, save_suffix)), path = nothing, save_prefix = nothing, save_suffix = nothing, save_delim = "_", ext = r"fits(\\.tar\\.gz)?"i, kwargs...)
+    arrays(f,
+           collection;
+           save = any(!isnothing, (save_prefix, path, save_suffix)),
+           path = nothing,
+           save_prefix = nothing,
+           save_suffix = nothing,
+           save_delim = "_",
+           ext = r"fits(\\.tar\\.gz)?"i,
+           kwargs...)
 
 Iterates over the image arrays of the collection applying function `f` at each step.
 
@@ -366,7 +406,7 @@ end
 The above generates `processed_images` which consists of trimmed versions of image arrays present in `collection`.
 For saving the `processed_images` simultaneously with the operations performed
 ```julia
-processed_images = map(arrays(collection; save = true, path = "~/data/tekdata", save_prefix = "trimmed")) do img
+processed_images = map(arrays(collection; path = "~/data/tekdata", save_prefix = "trimmed")) do img
     trim(img, (:, 1040:1059))
 end
 ```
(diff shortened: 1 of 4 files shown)


siddharthlal25:
I will rename `df` to `collection` at the end of the PR to keep the diff clean for reviewing.

src/collection.jl
```diff
@@ -9,9 +9,64 @@ end
 
 parse_name(filename, ext, ::Val{true}) = filename
 
+# utility function for generating filename
+function generate_filename(path, save_location, save_prefix, save_suffix, save_delim, ext)
+    # get the filename
+    filename = last(splitdir(path))
```
giordano: Isn't it equivalent to
```suggestion
    filename = basename(path)
```
?

siddharthlal25: Made the change, thank you!

siddharthlal25:
@mileslucas @giordano this one is ready for review! Let me know your thoughts about the code.

src/collection.jl
```diff
@@ -130,15 +225,167 @@ end
...
+end
+```
 """
-function images end
+function ccds end
 
-# generator for ImageHDU specified by data frame (i.e. path of file, hdu etc.)
-@resumable function images(df::DataFrame)
+# generator for CCDData specified by data frame (i.e. path of file, hdu etc.)
+@resumable function ccds(df::DataFrame)
```
mileslucas: I'm just thinking, do you need to specialize on `DataFrame` here? there's only one version of `ccds(x)` (the other is `ccds(f, x)`) and this way we can accept a much more expansive set of table types. For example, just because we return the collection as a DataFrame doesn't mean the user won't modify it or want to change it into a different type. As far as we are concerned, the only thing we need is the interface to `eachrow` and some property accessors. Theoretically the input could just be a vector of named tuples, if they're structured correctly!

src/collection.jl
```diff
@@ -130,15 +225,167 @@ end
...
+
+Mapping version of `ccds` function is interfaced on iterative version of `images`, any valid parameter can be passed into iterative version as `kwargs`.
+"""
+function ccds(f, collection::DataFrame; save = false, path = nothing, save_prefix = nothing, save_suffix = nothing, save_delim = "_", ext = r"fits(\.tar\.gz)?"i, kwargs...)
+    image_iterator = ccds(collection; kwargs...)
+    locations = collection.path
+
+    processed_images = map(zip(locations, image_iterator)) do (location, output)
+        processed_image = f(output)
+        if save
```
mileslucas, line 298: I'm torn here-

on one hand, it's _very_ clear and _very_ simple to only save via a keyword argument. However, consider `ccds(pipelin, coll; save_suffix="_cal")`. This won't actually save the files, and I now have to manually add `save=true` even though I may have assumed that it _would_ save!

I would lean towards some automatic `save` behavior that can be overridden instead of requiring it. Something like
```julia
save = any(!isnothing, (save_prefix, path, save_suffix))
```

src/fits.jl
```diff
@@ -1,11 +1,13 @@
...
-=#
+"""
+    getdata(::FITSIO.ImageHDU)
+
+Loads image array form `ImageHDU`
+
+FITSIO.jl takes over memory read in by cfitsio, which reads in row-major form, whereas when Julia takes that memory, it is assumed as column major.
+Therefore all data read by [`FITSIO.read`](http://juliaastro.github.io/FITSIO.jl/latest/api.html#Base.read-Tuple{ImageHDU}) is transposed. This function allows the user to read data in a consistent way to `Array` by transposing after reading.
+"""
 function getdata(hdu::ImageHDU)
```
mileslucas, line 11: `write_data` and `getdata` offers some grammatical inconsistency. I would settle on `getdata` (mimics an astropy function, is *getting* the *data* as an array from the fits file) and `writefits` (matches grammar of no underscore, is *writing* the data to a *fits* file)

little inconsistencies like this if we correct them now, hopefully we avoid burdening future developers and users!

src/collection.jl
siddharthlal25: > Theoretically the input could just be a vector of named tuples, if they're structured correctly!

I think the user would have to dig in a bit for putting up a correct vector of named tuples, and if the user doesn't do it correctly, it may lead to errors. That's why I thought of explicitly writing `DataFrame` in the description to avoid this confusion.

Although, only writing `DataFrame` would not make sense, let's modify the docs, and keep it as:
`df` can be a vector of named tuples or `DataFrame` with some specific properties.

Also, I am not sure about the compatibility of `eachrow` with vector of named tuples.

src/collection.jl
siddharthlal25: So, should we remove the `save` keyword from the list of input parameters?

PS: Keeping it would make sure that we will be saving (definitely sure!)
Whereas it is possible that a user may set `save_prefix` without actually wanting to save it.

PS: I don't think that makes a lot of sense.
Let's remove save from keywords, what do you say?

src/fits.jl
siddharthlal25: Sure, I will switch to `writefits`!

src/collection.jl
mileslucas: Actually, a user has to do extremely little to transfer between table types thanks to the Tables.jl interface. I'm away from my computer but I can easily take a dataframe from the collection function, turn it into a TypedTable from TypedTables.jl and there's no reason that shouldn't work with our function. So specializing in DataFrame means the users can't use anything else. My point is that we don't need the specialization AND it restricts our users, so why have it?

src/collection.jl
mileslucas: If you make save a keyword with a default value (like I just showed you) means that it will behave "automatically" but the user can still override it. Without the keyword argument the user can't override it.

src/collection.jl
siddharthlal25: Ya, that's a point. Let's drop it!

src/collection.jl
```diff
@@ -115,30 +194,198 @@ end
...
+        end
+        processed_image
+    end
+
+    return processed_images
+end
+
+
+"""
+    filenames(f, collection; save = any(!isnothing, (save_prefix, path, save_suffix)), path = nothing, save_prefix = nothing, save_suffix = nothing, save_delim = "_", ext = r"fits(\\.tar\\.gz)?"i, kwargs...)
```
mileslucas: maybe consider breaking this line up due to its length and re-ordering it so `save` appears _after_ the kwargs that it is based off of.

src/collection.jl
```diff
@@ -115,30 +194,198 @@ end
...
+
+Iterates over the file paths of the collection applying function `f` at each step.
+
+It returns an array of output values of function `f` applied on file paths.
```
```suggestion
Loads the data associated with each row of `collection` into an `Array` and applying `f` to the data.
```

src/collection.jl
```diff
@@ -115,30 +194,198 @@ end
...
+In addition to applying function `f`, the outputs can be saved. If `save = true`, it enables programmatical saving of returned value of the function `f` using [`CCDReduction.writefits`](@ref). File is saved at `path` specified by the user.
+Suffix and prefix can be added to filename of newly created files by modifying `save_suffix` and `save_prefix`, `save_delim` is used as delimiter.
+`ext` is the extension of files in collection, by default it is set to `r"fits(\\.tar\\.gz)?"i`.
```
```suggestion
The output from `f` can be saved using the appropriate keyword arguments. The `save_prefix` argument will add a prefix to each filename delimited by `save_delim`. `save_suffix` will add a suffix prior to the extension, which can be manually provided via `ext`, similar to [`fitscollection`](@ref). Files will be saved in the directory they are stored unless `path` is given. Finally, `save` will default to `true` if any of the previous arguments are set, but can be manually overridden (useful for testing). Files will be saved using [`CCDReduction.writefits`](@ref).
```

src/collection.jl
```diff
@@ -115,30 +194,198 @@ end
...
+    fh = FITS(path)
+    data = getdata(fh[1]) # assuming all 1-hdu are ImageHDUs
+    close(fh)
+    data
+end
+```
+The above generates `data` which consists of image arrays corresponding to 1st hdu of FITS file paths present in `collection`.
+For saving the `data` simultaneously with the operations performed
+```julia
+data = map(filenames(collection; save = true, path = "~/data/tekdata", save_prefix = "retrieved_from_filename")) do img
```
```suggestion
data = map(filenames(collection; path = "~/data/tekdata", save_prefix = "retrieved_from_filename")) do img
```

src/fits.jl
```diff
@@ -1,11 +1,13 @@
 # helper functions
 
-#=
-FITSIO.jl takes over memory read in by cfitsio, which reads in row-major form,
-whereas when Julia takes that memory, it is assumed as column major.
-Therefore all data read by `read` is transposed.
-Related comment: https://github.com/JuliaAstro/CCDReduction.jl/pull/16#issuecomment-638492572
-=#
+"""
+    getdata(::FITSIO.ImageHDU)
```
mileslucas: Is this exported? If not, I would change to `CCDReduction.getdata(::FITSIO.ImageHDU)`

src/fits.jl
```diff
@@ -1,11 +1,13 @@
...
+
+Loads image array form `ImageHDU`
```
mileslucas:
```suggestion
Loads the given HDU as an `Array`, permuting the dimensions appropriately.
```

src/collection.jl
siddharthlal25: Shouldn't we mention that it *just* gives the `filenames`?

> Loads the data associated with each row of `collection` into an `Array`

PS: This sounds like we are loading the images in an array

src/collection.jl
siddharthlal25: > Files will be saved in the directory they are stored unless `path` is given

Should we implement this? Or should we leave this for user? (Haven't implemented this yet, as in rn one has to explicity mention the path)

PS: I don't think it would be a trouble for the user to go for `path = @__DIR__`
Thoughts?

mileslucas: `@__DIR__` would be the same directory as where this file is, that is definitely not what we want. I think the two options we have are either the original path or the current working directory. I think the one that makes more sense  is to have path default to the same directory as the file

src/collection.jl
mileslucas: Ah, I was mistaken, I though based off the previous docstrings that this was `arrays`. Yes it should just say it returns the filenames, which wasn't clear before.

siddharthlal25: > Yes it should just say it returns the filename

This will not always return filenames, I mean it can return `CCDData` if we do:
```
filenames(collection) do name
    CCDData(name)
end
```

mileslucas: careful, that's not what `filenames` returns. `filenames` actually returns an iterator of strings, roughly. The strings are the filenames. In the example you showed, the return value is a consequence of using this iterator, but the iterator itself is still over the filenames. Hopefully that makes some sense.

src/collection.jl
```diff
@@ -115,30 +194,252 @@ end
...
 end
 
 
 """
-    images(df::DataFrame)
+    ccds(collection)
+
+Generator for `CCDData`s of entries in data frame.
+
+Iterates over `collection` using each `path` and `hdu` to load data using [`FITSIO.FITS`](http://juliaastro.github.io/FITSIO.jl/latest/api.html#FITSIO.FITS) into a [`CCDData`](@ref).
```
```suggestion
Iterates over `collection` using each `path` and `hdu` to load data into a [`CCDData`](@ref).
```
I don't seen the need to include that extra detail at the sacrifice of grammatical flow

src/collection.jl
```diff
@@ -115,30 +194,252 @@ end
...
+
+For saving the `processed_images` simultaneously with the operations performed
+```julia
+processed_images = map(ccds(collection; path = "~/data/tekdata", save_prefix = "trimmed")) do img
+    trim(img, (:, 1040:1059))
+end
+```
+The trimmed images are saved as `trimmed_(original_name)` (FITS files) at `path = "~/data/tekdata"` as specified by the user.
+
+Mapping version of `ccds` function is interfaced on iterative version of `images`, any valid parameter can be passed into iterative version as `kwargs`.
```
```suggestion
```
I don't think we need to explicitly state this

src/collection.jl
siddharthlal25: What would be a nice way of mentioning this in docs? i.e. it returns filenames but at the end it actually outputs an `Array` of `f` (function) operated on each filename.

mileslucas: Well, that's still not true. Imagine the user creates a function that returns nothing. The output will then be a bunch of nothings! The output is in the users hands, so we shouldn't mention what the output is, or else we stand a chance of being wrong. All we can (and should) say is that it will apply the user function to the filename at each iteration

src/collection.jl
```diff
@@ -115,30 +194,252 @@ end
...
+                   save = any(!isnothing, (save_prefix, path, save_suffix)),
+                   save_delim = "_",
+                   ext = r"fits(\.tar\.gz)?"i,
+                   kwargs...)
+    path_iterator = filenames(collection; kwargs...)
+    locations = collection.path
+
+    processed_images = map(zip(locations, path_iterator)) do (location, output)
+        processed_image = f(output)
+        if save
```
mileslucas, line 364: I've just realized we should throw a check here, too (for all 3 versions)
`if save && !isnothing(processed_image)`

src/collection.jl
siddharthlal25: ~~Perfect, I will modify the docs!~~

Edit:

I agree that user can make a function that returns `nothing` and finally the array returned will contain `nothing`, but this does not make the point incorrect, because at the end we will still have an array of output values of `f` (containg nothing as mentioned in docs, i.e. it returns the output).

Thoughts?

src/collection.jl
siddharthlal25: Shouldn't this be:
`if save && (processed_image isa Array || processed_image isa CCDData)`?
Because we can only save `Array` and `CCDData`.

mileslucas: I don't think we should use an if-statement to enforce type checking. We should let multiple dispatch handle that. The reasoning behind checking for `nothing` is functions *that don't have a return value*.

Thinking back on it, we probably shouldn't do this, since it would be weird to try saving something and failing silently- the user wouldn't see any error or warning that the data isn't being saved. If an error gets thrown (a method error for writefits) that'll fail loudly.

src/collection.jl
mileslucas: The point I'm trying to make is that all we need to describe is what makes this different from `arrays` and `ccds`. So instead of saying "this returns a vector of arrays" which is talking about the output of the user's own function, we should emphasize that this iterates and provides filenames. Anything beyond that is related to the user's function, and I don't think we should try writing documentation based on something we can't predict

siddharthlal25:
I have made a small change, I have exported the functions `data` and `hdr` to be accessible by the user for convenience and uniform interface. Thoughts?

src/collection.jl
siddharthlal25: Pushed the changes!

siddharthlal25:
I am merging this one! Thanks for the help and reviews!

PR #47, opened 2024-11-24
Compat bump for ResumableFunctions and a versionbump for release

Krastanov:
ResumableFunctions 1.0.0 was released, now with proper support for variable scopes (previously it was breaking local scopes). I tested it locally and it runs with CCDReduction.jl, so providing a version bump here.

CCDReduction.jl is also used in the CI of ResumableFunctions to check for unintentional breaking changes. Having a new release of CCDReduction.jl would be valuable for our CI.


cgarling:
Seems fine to me but I am not very familiar with this codebase. I will wait to merge until the end of the week in case any more qualified maintainers wish to comment.

